PR #4, opened 2019-08-01
Some documentation cleanups

ethomson:
* Only specify `-f` (force) once.
* Submodule documentation improvements

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -11,7 +11,7 @@ inputs:
     description: 'If true, execute `execute git clean -ffdx && git reset --hard HEAD` before fetching'
     default: true
   submodules:
-    description: 'Directory containing files to upload'
+    description: 'Whether to recursively clone submodules; defaults to false'
   lfs:
     description: 'Whether to download Git-LFS files; defaults to false'
   fetch-depth:


kivikakk:
I did some empirical testing (because I wanted submodules!):

* [without `submodules: true`](https://github.com/kivikakk/comrak/runs/191881491)
* [with `submodules: true`](https://github.com/kivikakk/comrak/pull/128/checks?check_run_id=191883812)

The difference is, with `submodules: true` the following get executed at the end of the action:

```
git submodule sync
git -c http.https://github.com.extraheader="AUTHORIZATION: basic ***" submodule update --init --force
```

So I'd say the proposed improvement is accurate!

TingluoHuang:
@ethomson we do run `git clean -ffdx` to delete nested git repo under the main repo.
```
If the Git configuration variable clean.requireForce is not set to false, git clean will refuse to delete files or directories unless given -f, -n or -i. Git will refuse to delete directories with .git sub directory or file unless a second -f is given.
```

ethomson:
Thanks @TingluoHuang, I checked the docs again and you're obviously right.

Issue #20, opened 2019-08-24
How to checkout the current branch?


zeke:
I want to set up a job that checks out the branch that was pushed. Is it possible to do something like this?

```yml
- uses: actions/checkout@${GITHUB_REF}
```

Reference: https://help.github.com/en/articles/virtual-environments-for-github-actions#environment-variables

TingluoHuang:
try:
```
- uses: actions/checkout@master
  with:
    ref: ${{ github.ref }}
```

TingluoHuang:
BTW, if you just do `- uses: actions/checkout@master`, it will checkout the `GITHUB_SHA`, the `@master` is for versioning this `checkout` action

zeke:
Ah thanks! Should I open a PR to add something to the README for this use case?

TingluoHuang:
sure.

zeke:
Opened a PR: https://github.com/actions/checkout/pull/21

fbartho:
PR is already merged, so this issue could be closed!

ethomson:
Thanks, @zeke!

PR #21, opened 2019-08-24
add docs about checking out the triggering ref

zeke:
See #20

cc @TingluoHuang

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ See [here](https://help.github.com/en/articles/events-that-trigger-workflows) to
 See [action.yml](action.yml)
 
 Basic:
+
 ```yaml
 steps:
 - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +21,16 @@ steps:
 - run: npm test
 ```
 
+By default, the branch or tag ref that triggered the workflow will be checked out. If you wish to check out a different branch, specify that using `with.ref`:
```
jasonkarns: This isn't quite right. The branch/tag's HEAD will be checked out, which means (by default) the repo is in a Detached HEAD state. That's different than checking out the branch or tag, which is _not_ a detached HEAD.

zeke: Thanks for the clarification. Do you want to open a followup PR to make these docs more clear?

sfc-gh-zhwang: how to checkout to the branch without bing detached?

zeke: @sfc-gh-zhwang you should be able to fork the repo, clone it, and work off the main branch.

PR #31, opened 2019-09-06
Add explanatory comment to plugin

damccorm:
@thboop to review. Once this is in we should do the same in upload/download artifact


damccorm:
See https://github.com/actions/toolkit/issues/69 for context

Issue #13, opened 2019-08-16
Set git user and email


wesleytodd:
It would be great if this action would also set the `email` and `name`.  In order to do other git operations it is required.  I am adding this to my actions to get around it:

```yaml
steps:
  - uses: actions/checkout@v1
  - name: Setup git user
    run: |
      git config user.email "<>"
      git config user.name "${GITHUB_ACTOR}"
```

Unfortunately I don't have access to the user email. It seems like these can come from the account in a more clean way here, but I guess I could also create my own action.  Since this is so new I am not sure what the best way forward is so thought I would post here.

TingluoHuang:
@chrispat for feedback.

1138-4EB:
This would be useful for users willing to update `gh-pages` hosted on separate repos using a deployment key (e.g. https://github.com/1138-4EB/vunit/blob/migrate-to-gha/.github/workflows/push.yml#L63-L74). This is specially so, because `GITHUB_TOKEN`'s scope is limited and personal access tokens are not scoped to a single target repo.

wesleytodd:
> This would be useful for users willing to update gh-pages

This was actually my first use case for this!

https://github.com/pkgjs/gh-pages/blob/master/entrypoint.sh#L4-L5

fregante:
I recently wrote an action that sets up the user, email and token in one go, so you can also push back to the repo.

https://github.com/fregante/setup-git-token

```yml
    - uses: fregante/setup-git-token@v1
      with:
        token: ${{ secrets.GITHUB_TOKEN }}
        name: The Bot
        email: bot@example.com
```

eine:
@ericsciple, should this issue be closed now that #107 is merged? It seems to be included in v2: https://github.com/actions/checkout/commits/v2

wesleytodd:
So that PR is a bit hard to grok on my phone (even with the improvements in the GitHub app), but it is unclear if that sets the `name` and `email`. This issue is specifically about that need. Does that PR solve this use case indirectly?

eine:
@wesleytodd, see:

https://github.com/actions/checkout/blob/3537747199ad29df25693bc607e99df5d7726ffd/action.yml#L12-L17

It seems to me that this issue was/is about "*do other (authenticated) git operations*"; precisely update gh-pages (which was/is your "*first use case for this*"). Explicitly setting the user and email would have been an approach. Using the token to do so is just another suitable solution.

[...]

PR #56, opened 2019-10-14
Upgrade checkout to v1.1.0 to better support scripting git.

TingluoHuang:
https://github.com/github/dreamlifter/issues/1297

Checkout will persist the git credential into the repository git config file.

If the token has write-access to the repository, workflow user can use
```yaml
- run: |
     git commit -am "changes"
     git push
```
to push the change back to the repository without dealing with git authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,8 @@
-# checkout
+<p align="center">
+  <a href="https://github.com/actions/checkout"><img alt="GitHub Actions status" src="https://github.com/actions/checkout/workflows/test-local/badge.svg"></a>
+</p>
+
+# Checkout
 
 This action checks out your repository to `$GITHUB_WORKSPACE`, so that your workflow can access the contents of your repository.
 
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ Basic:
 
 ```yaml
 steps:
-- uses: actions/checkout@v1.0.0
-- uses: actions/setup-node@master
+- uses: actions/checkout@v1
+- uses: actions/setup-node@v1
   with:
     node-version: 10.x 
 - run: npm install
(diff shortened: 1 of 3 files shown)


peter-evans:
I'm very interested to know how this is implemented.

- What will be the default `user.name` and `user.email` set in git config? Will that vary depending on the event type?
- Presumably, for this to work, a branch needs to be checked out instead of leaving the repository in a detached HEAD state. For most event types I'm guessing `refs/heads/<branch>` will be checked out. What will happen for `on: pull_request` workflows? In `checkout@v1` it creates a merge commit. To avoid that commit being pushed I'm checking out `github.head_ref` which destroys the merge commit.

README.md
```diff
@@ -13,23 +13,48 @@ Basic:
 
 ```yaml
 steps:
-- uses: actions/checkout@master
+- uses: actions/checkout@v1.0.0
```
bryanmacfarlane: No, we should not suggest this as the mainline (README).  the mainline is @v1 and binding to a specific patch release is a way to mitigate an issue and bind back if "latest v1" breaks you.  Actions `v1` is equal to "1.x".

thboop: I addressed this feedback

thboop:
Hey @peter-evans!
> I'm very interested to know how this is implemented.

We are working on making this more transparent! 

> * What will be the default `user.name` and `user.email` set in git config? Will that vary depending on the event type?

There's an open feedback ticket [here](https://github.com/actions/checkout/issues/13) for that functionality, we'd love for you to chime in on what would help you build the best workflows.!

To clarify, this pr sets neither the email or name, in order to push you will need to set those using the appropriate git commands

> * ~For most event types I'm guessing `refs/heads/<branch>` will be checked out. What will happen for `on: pull_request` workflows? In `checkout@v1` it creates a merge commit. To avoid that commit being pushed I'm checking out `github.head_ref` which destroys the merge commit.~ After thinking about this a bit more I guess the behaviour won't change. I will still need to checkout `github.head_ref` instead of the merge commit.

We haven't made any changes to what is checked out, but it should be a little easier for you to checkout or fetch something you may need without doing extra work to auth!

randydwhitephotography (commented):
ugrade

randydwhitephotography (commented):
signed-off-by Randy D White randydwhite@randydwhite.com signed-off-by Randy D White randydwhite@randydwhite.com

randydwhitephotography (commented):
signed-off-by Randy D White randydwhite@randydwhite.com

PR #70, opened 2019-11-21
checkout v2

ericsciple:


.github/workflows/test.yml
```diff
@@ -1,19 +1,59 @@
-name: "test-local"
+name: Build and Test
+
 on:
   pull_request:
   push:
     branches:
       - master
       - 'releases/*'
+      - users/ericsciple/action
```
ericsciple: remove this

.github/workflows/test.yml
```diff
@@ -1,19 +1,59 @@
...
+      - uses: actions/checkout@v1 # todo: switch to v2
+      - run: npm install
+      - run: npm run build
+
   test:
-    strategy:
-      matrix:
-        os: [windows-latest, ubuntu-latest, macOS-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
```
ericsciple: test on linux macos and windows

.github/workflows/test.yml
```diff
@@ -1,19 +1,59 @@
...
+        uses: ./
+        with:
+          repository: ericsciple/testing # switch to actions/checkout, this must be hardcoded doesnt work in a fork
+          ref: test-data/v2/lfs
+          path: lfs
+          lfs: true
+      - name: Verify LFS
+        run: src/test/verify-lfs.sh
+
+      # Submodules false
```
ericsciple: change the wording. basically this verifies the submodule is not checked out

lib/checkout.js
```diff
@@ -0,0 +1,120 @@
+"use strict";
```
ericsciple: ignore the js, review the ts further below

node_modules/.bin/semver
```diff
@@ -0,0 +1 @@
+../semver/bin/semver
```
ericsciple: i think i might need to run ncc, havent looked into it yet

src/checkout.ts
```diff
@@ -0,0 +1,120 @@
+import * as core from '@actions/core';
+import * as coreCommand from '@actions/core/lib/command';
+import * as fsHelper from './fs-helper';
+import * as gitSourceProvider from './git-source-provider';
+import * as github from '@actions/github'
+import * as path from 'path';
+
+async function run() {
```
ericsciple: this is the entry file and it just gets and validates the inputs

src/checkout.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        // Access token
+        let accessToken = core.getInput('token');
+
+        try {
+            // Register problem matcher
+            coreCommand.issueCommand('add-matcher', {}, path.join(__dirname, 'problem-matcher.json'));
+
+            // Get sources
+            await gitSourceProvider.getSource(
```
ericsciple: call the source provider (effectively the business logic layer)

src/fs-helper.ts
```diff
@@ -0,0 +1,80 @@
+import * as fs from 'fs';
+
+export function directoryExistsSync(
```
ericsciple: some useful helpers i think we should add to the toolkit.... separate pr

src/git-command-manager.ts
```diff
@@ -0,0 +1,399 @@
+import * as core from '@actions/core'
+import * as exec from '@actions/exec'
+import * as fshelper from './fs-helper';
+import * as io from '@actions/io';
+import * as path from 'path';
+
+export interface IGitCommandManager {
```
ericsciple: this is a wrapper over git.exe and exposes methods that correlate 1-to-1 with the git commands we need to run.

src/git-command-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+    remoteAdd(remoteName: string, remoteUrl: string): Promise<void>;
+    tagExists(pattern: string): Promise<boolean>;
+    tryClean(): Promise<boolean>;
+    tryConfigUnset(configKey: string): Promise<boolean>;
+    tryDisableAutomaticGarbageCollection(): Promise<boolean>;
+    tryGetFetchUrl(): Promise<string>;
+    tryReset(): Promise<boolean>;
+}
+
+export async function CreateCommandManager(
```
ericsciple: we need to do some async stuff when initializing the GitCommandManager class, so this function is exposed and the ctor for GitCommandManager is private

src/git-source-provider.ts
```diff
@@ -0,0 +1,292 @@
...
+import * as coreCommand from '@actions/core/lib/command';
+import * as fs from 'fs';
+import * as fsHelper from './fs-helper';
+import * as gitCommandManager from './git-command-manager';
+import * as io from '@actions/io';
+import * as path from 'path';
+
+let authConfigKey = `http.https://github.com/.extraheader`;
+
+export async function getSource(
```
ericsciple: here is the source provider. this basically drives the business logic

src/git-source-provider.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    // Dump some info about the checked out commit
+    await git.log1();
+
+    // Set intra-task state for cleanup
+    coreCommand.issueCommand('save-state', { name: 'repositoryPath' }, repositoryPath);
+    coreCommand.issueCommand('save-state', { name: 'configKey' }, authConfigKey);
+};
+
+export async function cleanup() {
```
ericsciple: i need to revisit this and make sure its best effort (doesnt fail the step if the repo isnt there anymore or whatever)

src/post-checkout.ts
```diff
@@ -0,0 +1,13 @@
+import * as core from '@actions/core';
+import * as gitSourceProvider from './git-source-provider';
+
+async function run() {
```
ericsciple: entry point for the POST step

lib/problem-matcher.json
```diff
@@ -0,0 +1,13 @@
...
+            "owner": "checkout-git",
+            "pattern": [
+                {
+                    "regexp": "^(fatal|error): (.*)$",
+                    "message": 2
+                }
+            ]
+        }
+    ]
+}
```
ericsciple: skip the node_modules, jump down further

ericsciple: investigating ncc

src/git-source-provider.ts
```diff
@@ -0,0 +1,285 @@
...
+import * as path from 'path';
+
+let authConfigKey = `http.https://github.com/.extraheader`;
+
+export async function getSource(
+    repositoryPath: string,
+    repositoryOwner: string,
+    repositoryName: string,
+    ref: string,
+    commit: string,
```
hross: I would make all of these arguments in the function into an interface and pass this as a single object to keep this simple.

src/git-source-provider.ts
```diff
@@ -0,0 +1,285 @@
+import * as core from '@actions/core';
+import * as coreCommand from '@actions/core/lib/command';
+import * as fs from 'fs';
+import * as fsHelper from './fs-helper';
+import * as gitCommandManager from './git-command-manager';
```
hross: I would import the individual classes/interfaces if you are using 2 or less (in this case you are using the type a lot so it would make it simpler to declare variables)

src/git-source-provider.ts
```diff
@@ -0,0 +1,285 @@
...
+    repositoryName: string,
+    ref: string,
+    commit: string,
+    clean: boolean,
+    fetchDepth: number,
+    lfs: boolean,
+    accessToken: string) {
+
+    core.info(`Syncing repository: ${repositoryOwner}/${repositoryName}`);
+    let repositoryUrl = `https://github.com/${encodeURIComponent(repositoryOwner)}/${encodeURIComponent(repositoryName)}`;
```
hross: should use const here if not redefining this later, explicitly define it as : string

src/git-source-provider.ts
```diff
@@ -0,0 +1,285 @@
...
+        isNew = true;
+        await io.mkdirP(repositoryPath);
+    }
+
+    // Git command manager
+    core.info(`Working directory is '${repositoryPath}'`);
+    let git = await gitCommandManager.CreateCommandManager(repositoryPath, lfs);
+
+    // Existing directory
+    if (!isNew) {
```
hross: I would take this whole if block and make a method since the parent method is really large and complicated. You can make the return value from this be equal to recreate, which lets you short-circuit all the if blocks and directly return from the method

src/git-source-provider.ts
```diff
@@ -0,0 +1,285 @@
...
+            recreate = true;
+        }
+        // Fetch URL matches
+        else {
+            // Delete any index.lock and shallow.lock left by a previously canceled run or crashed git process
+            let lockPaths = [
+                path.join(repositoryPath, '.git', 'index.lock'),
+                path.join(repositoryPath, '.git', 'shallow.lock')
+            ];
+            for (let i = 0; i < lockPaths.length; i++) {
```
hross: would use for of here since you don't need the index variable and use const for the variable in the loop

src/git-source-provider.ts
```diff
@@ -0,0 +1,285 @@
...
+        let upperConfigKey = configKey.toUpperCase();
+        let split = contents.split('\n').filter(x => x.toUpperCase().indexOf(upperConfigKey) < 0);
+        contents = split.join('\n');
+
+        // Rewrite the config file
+        fs.writeFileSync(configPath, contents);
+    }
+}
+
+interface ICheckoutInfo {
```
hross: put interfaces at the top of the file, or maybe more ideal make it a separate file unless it's part of a class

src/fs-helper.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (!path) {
+        throw new Error("Arg 'path' must not be empty");
+    }
+
+    let stats: fs.Stats;
+    try {
+        stats = fs.statSync(path);
+    }
+    catch (error) {
+        if (error.code == 'ENOENT') {
```
hross: use === here

src/fs-helper.ts
```diff
@@ -0,0 +1,80 @@
...
```
hross: ===

src/git-command-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+        configKey: string,
+        configValue: string) {
+
+        await this.execGit(['config', configKey, configValue]);
+    }
+
+    public async configExists(configKey: string): Promise<boolean> {
+        let pattern = configKey.replace(/[^a-zA-Z0-9_]/g, (x) => { return `\\${x}` });
+        let output = await this.execGit(['config', '--name-only', '--get-regexp', pattern], true);
+        return output.exitCode == 0;
```
hross: ===

src/git-command-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+            throw new Error("Arg minimum is not a valid version");
+        }
+
+        // Major is insufficient
+        if (this.major < minimum.major) {
+            return false;
+        }
+
+        // Major is equal
+        if (this.major == minimum.major) {
```
hross: === (i'll stop adding === comments but you should always use === or !== unless you expect a truthy return value

src/checkout.ts
```diff
@@ -0,0 +1,149 @@
+import * as core from '@actions/core';
+import * as coreCommand from '@actions/core/lib/command';
+import * as fsHelper from './fs-helper';
+import * as gitSourceProvider from './git-source-provider';
+import * as github from '@actions/github'
+import * as path from 'path';
+
+async function run() {
+    try {
```
hross: add comment here about using error to break out of the method since there are methods below with side effects like error conditions that we expect to happen and stop continuation (and that is fully okay)

src/git-command-manager.ts
```diff
@@ -0,0 +1,399 @@
+import * as core from '@actions/core'
```
hross: ideally we would use jest or some kind of testing framework to write unit tests around these classes, rather than just the shell scripts. That would help other people write more tests in the future and make the tests more bite sized to the JS classes, plus define expected input/output

tsconfig.json
```diff
@@ -0,0 +1,66 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
```
hross: should we just remove all the commented lines? seems weird to have them all in here

.eslintrc.json
```diff
@@ -0,0 +1,58 @@
+{
```
ericsciple: copied from actions/toolkit

.github/workflows/test.yml
```diff
@@ -1,19 +1,59 @@
...
       - 'releases/*'
+      - users/ericsciple/action
 
 jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1 # todo: switch to v2
+      - run: npm install
+      - run: npm run build
```
ericsciple: also run check-format, lint, and pack. after pack fail if any edits from git

.prettierrc.json
```diff
@@ -0,0 +1,11 @@
+{
```
ericsciple: copied from actions toolkit

tsconfig.json
ericsciple: i think i copied this from somewhere or it was generated. it was at least a few weeks ago so i dont remember.

i dont have a strong preference either way. its just the way it was (generated or copied) and i didnt see any value in changing it

src/git-source-provider.ts
ericsciple: currently its only used privately within this file. i can move to a separate file but i'm not sure that helps readability?

src/git-command-manager.ts
ericsciple: +1 there is a couple places in this module with logic that would be interesting to unit test

most of the logic is uninteresting though. biggest value for investment simple e2e w/ a couple shell scripts to assert expected layout

.github/workflows/test.yml
```diff
@@ -1,19 +1,64 @@
-name: "test-local"
+name: Build and Test
+
 on:
   pull_request:
   push:
     branches:
       - master
       - 'releases/*'
+      - users/ericsciple/action
```
ericsciple: remove this after done testing

src/checkout.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    try {
+      // Register problem matcher
+      coreCommand.issueCommand(
+        'add-matcher',
+        {},
+        path.join(__dirname, 'problem-matcher.json')
+      )
+
+      // Get sources
```
ericsciple: add `persist-credentials` input

src/git-source-provider.ts
ericsciple: rationale being - contained within this module privately is advantageous that it doesnt bloat number of files. it's a tradeoff either way.

ericsciple: btw i added eslint and prettier config same as used in the toolkit. should cut down on most style issues

tsconfig.json
ericsciple: i think this is another place where it's a tradeoff eitherway. tsc saw value in generating the file this way for readers.

ericsciple: you could argue now that the decisions are made we dont need the comments anymore. seems reasonable. i'll remove them

ericsciple: btw i think i copied this from setup-node

src/fs-helper.ts
```diff
@@ -0,0 +1,77 @@
+import * as fs from 'fs'
```
ericsciple: some useful helpers i think we should add to the toolkit.... separate pr

src/main.ts
```diff
@@ -0,0 +1,163 @@
+import * as core from '@actions/core'
+import * as coreCommand from '@actions/core/lib/command'
+import * as fsHelper from './fs-helper'
+import * as gitSourceProvider from './git-source-provider'
+import * as github from '@actions/github'
+import * as path from 'path'
+
+async function run(): Promise<void> {
```
ericsciple: this is the entry file. it just validates the inputs, and then calls the source provider which performs all of the business logic

src/git-command-manager.ts
ericsciple: added eslint/prettier

src/fs-helper.ts
ericsciple: added eslint/prettier

src/fs-helper.ts
ericsciple: added eslint/prettier

src/git-source-provider.ts
ericsciple: added eslint/prettier

src/git-source-provider.ts
ericsciple: added prettier/eslint

src/git-source-provider.ts
ericsciple: personally not a fan of that pattern. i find the indirection hinders readability (now i have to find the class to see what the params are)

ericsciple: i added eslint/prettier to hopefully cut down on a lot of the style stuff

src/git-source-provider.ts
ericsciple: +1 it reads better now

src/checkout.ts
ericsciple: i dont understand what the concern is?

note, i renamed the file to main.ts so it should be more clear this is the outer try/catch

src/git-command-manager.ts
ericsciple: chasing down other details right now, will try to circle back and add jest

src/git-command-manager.ts
ericsciple: added eslint/prettier

src/git-source-provider.ts
ericsciple: actually the pattern may come in handy after i get jest setup... will see...

README.md
```diff
@@ -2,60 +2,70 @@
...
-# Usage
+Refer [here](https://help.github.com/en/articles/events-that-trigger-workflows) to learn which commit `$GITHUB_SHA` points to for different events.
 
-See [action.yml](action.yml)
+Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous versions.
 
-Basic:
+# Usage
 
+<!-- start usage -->
```
ericsciple: i added a script to regenerate this from the action.yml

the CI verifies the docs are up to date

src/git-command-manager.ts
ericsciple: I added jest and refactored some of the interesting logic out into separate modules so i could isolate them via unit tests.

note, i intentionally did not invest in jest tests that make heavy use of mocking (high investment cost, low return). Rather I pulled interesting logic out into testable modules. Otherwise i'm leaning on the E2E coverage (easy to write, doesnt need to cover every scenario but pretty decent coverage with just a few e2e tests)

src/git-source-provider.ts
ericsciple: +1 i ended up using this pattern for classes/interfaces, not for exported functions though.

src/git-source-provider.ts
ericsciple: i moved this into a separate file along with a couple of the functions when i refactored for unit testing (public now from the helper module, moved to the top)

src/misc/generate-docs.ts
```diff
@@ -0,0 +1,102 @@
...
+import * as os from 'os'
+import * as path from 'path'
+import * as yaml from 'js-yaml'
+
+//
+// SUMMARY
+//
+// This script rebuilds the usage section in the README.md to be consistent with the action.yml
+
+function updateUsage(
```
ericsciple: we may want to move this function into a toolkit module

i added this based on feedback that the README.md should contain the inputs.

I didnt want to have to deal with maintaining both places, so now the CI runs the `npm run gendocs` and fails if it isnt up to date

src/git-version.ts
```diff
@@ -0,0 +1,63 @@
...
+        this.major = Number(match[1])
+        this.minor = Number(match[2])
+        if (match[4]) {
+          this.patch = Number(match[4])
+        }
+      }
+    }
+  }
+
+  checkMinimum(minimum: GitVersion): boolean {
```
hross: I would add access modifiers to these methods to make them clear

src/git-version.ts
```diff
@@ -0,0 +1,63 @@
+export class GitVersion {
```
hross: add comments to exported classes so consumers know what they do

src/git-source-provider.ts
```diff
@@ -0,0 +1,209 @@
...
+  repositoryOwner: string,
+  repositoryName: string,
+  ref: string,
+  commit: string,
+  clean: boolean,
+  fetchDepth: number,
+  lfs: boolean,
+  accessToken: string
+): Promise<void> {
+  core.info(`Syncing repository: ${repositoryOwner}/${repositoryName}`)
```
hross: you're missing a lot of line ending semicolons. not sure if this is intentional or not

src/git-source-provider.ts
ericsciple: done. i moved the args behind an interface. (lined-up with moving the input collection to a separate module for testability)

src/git-source-provider.ts
ericsciple: that's the prettier config. personally i think its weird but i just copied the config from the toolkit 🤷‍♂

src/git-version.ts
ericsciple: +1 however the prettier config didnt like it.

src/main.ts
```diff
@@ -0,0 +1,47 @@
...
+  try {
+    await gitSourceProvider.cleanup(cleanupRepositoryPath)
+  } catch (error) {
+    core.warning(error.message)
+  }
+}
+
+// Main
+if (!cleanupRepositoryPath) {
+  run()
```
TingluoHuang: do we need to `await run()`? (i am not good at the typescript async/await)

src/input-helper.ts
```diff
@@ -0,0 +1,104 @@
+import * as core from '@actions/core'
+import * as fsHelper from './fs-helper'
+import * as github from '@actions/github'
+import * as path from 'path'
+import {ISourceSettings} from './git-source-provider'
+
+export function getInputs(): ISourceSettings {
+  const result = ({} as unknown) as ISourceSettings
```
TingluoHuang: will `{} as ISouceSettings` work?

src/git-source-provider.ts
```diff
@@ -0,0 +1,222 @@
...
+    settings.repositoryOwner
+  )}/${encodeURIComponent(settings.repositoryName)}`
+
+  // Remove conflicting file path
+  if (fsHelper.fileExistsSync(settings.repositoryPath)) {
+    await io.rmRF(settings.repositoryPath)
+  }
+
+  // Create directory
+  let isExisting = true
```
TingluoHuang: do we need `isExisting` since we always `rm -rf` the `settings.repoPath`?

TingluoHuang: or we shouldn't always delete the local repo?

src/git-source-provider.ts
```diff
@@ -0,0 +1,222 @@
...
+    `x-access-token:${settings.accessToken}`,
+    'utf8'
+  ).toString('base64')
+  core.setSecret(base64Credentials)
+  const authConfigValue = `AUTHORIZATION: basic ${base64Credentials}`
+  await git.config(authConfigKey, authConfigValue)
+
+  // LFS install
+  if (settings.lfs) {
+    await git.lfsInstall()
```
TingluoHuang: do we need to do something similar to https://github.com/actions/runner/blob/master/src/Runner.Plugins/Repository/v1.0/GitSourceProvider.cs#L146

src/git-command-manager.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    const stdout: string[] = []
+
+    const options = {
+      cwd: this.workingDirectory,
+      env,
+      ignoreReturnCode: allowAllExitCodes,
+      listeners: {
+        stdout: (data: Buffer) => {
```
TingluoHuang: what about STDERR? should we also capture those?

src/git-command-manager.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+    if (!gitVersion.isValid()) {
+      throw new Error('Unable to determine git version')
+    }
+
+    // Minimum git version
+    // Note:
+    // - Auth header not supported before 2.9
+    // - Wire protocol v2 not supported before 2.18
+    const minimumGitVersion = new GitVersion('2.18')
```
TingluoHuang: double check to make sure MMS has this version on all hosted images.

src/main.ts
ericsciple: nope because we're not in an async function. node will wait on it.

src/input-helper.ts
ericsciple: unfortunately no either the linting or style rules we have doesnt like it

src/git-source-provider.ts
ericsciple: we dont always rmrf the repo path

src/git-source-provider.ts
ericsciple: it's in git-command-manager.ts in the initializeCommandManager function

src/git-command-manager.ts
ericsciple: it gets printed to the console but we dont use it to analyze the output

in other words, for example if we call `git branch --list` then stderr will flow to console.log, but we dont need to capture for the purpose of analyzing the branch names

TingluoHuang:
LGTM 👍

src/git-command-manager.ts
ericsciple: good idea. i checked all `-latest` and we're good

.github/workflows/test.yml
cm8972: > remove this

PR #90, opened 2019-12-03
document scenario: checkout the head commit of a PR

ericsciple:
related #61

PR #92, opened 2019-12-04
Update V2 Beta Docs

thboop:


README.md
```diff
@@ -3,6 +3,7 @@
 </p>
 
 # Checkout V2 beta
+The Checkout V2 beta is now available! You have automatically upgraded if you bound to the `master` branch. If you did not mean to upgrade, please bind to `uses: actions/checkout@v1`. Please see the [action versioning docs](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md) for more information.
```
ericsciple: Do we need to do this since the examples all reference `@v2-beta`?

This makes me feel like the content is leading with a technical note, rather than substance about the value of checkout v2.

Personal preference, I lean lean toward concise docs. I aim for skimmability (not a word, is now)

README.md
```diff
@@ -66,15 +67,15 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
 ## Checkout a different branch
 
 ```yaml
-- uses: actions/checkout@preview
+- uses: actions/checkout@v2-beta
```
ericsciple: i also have a pr to fix this: https://github.com/actions/checkout/pull/90

many users seem to be interested in checking out the HEAD commit of a PR rather than the merge commit. Seemed like an example was warranted (since it's otherwise not intuitive how to achieve it)

README.md
thboop: I viewed it as a short term note just to help those users who auto-upgraded on accident and came to this repo to see what happened, but I get your point. I'll close this pr.

Issue #61, opened 2019-10-29
Cannot use parent specifiers in the ref


mathstuf:
I'd like to run an action on a pull_request and check out the actual content of the PR, not a merge with `master` (because I want it to run even if there is a merge conflict among other reasons). Support for specifying this has been requested [on the community forum](https://github.community/t5/GitHub-Actions/Support-pull-request-events-without-merge-commit-into-the-target/td-p/36190).

I've tried the following:

```yaml
with:
  ref: ${{ github.ref }}^2
```

Which causes a failure because it tries to do:

```
 git fetch --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/pull/3/merge^2:refs/remotes/pull/3/merge^2
```

which is an invalid refspec.

Doing this seems to just get completely ignored:

```yaml
with:
  ref: ${{ github.sha }}^2
```

I haven't found a way to use `$GITHUB_SHA` yet either (it seems to be used literally when in a `with` block).

ericsciple:
@mathstuf consider adding a `run` step after checkout, and `git reset --hard HEAD^2`

ericsciple:
Yes. GITHUB_SHA is not set by the checkout action. It is from the server event payload.

mathstuf:
And while that may be a workaround, not being able to specify a valid Git refname in such a way (and that gets silently ignored) seems like a bug in any case.

mathstuf:
Also of note is that manual checkout misses all of the submodule sync/update, depth options, etc. available through the action.

ericsciple:
from the pr event payload, i think `${{ github.event.after }}` is the SHA you want

peter-evans:
I think this will checkout the merging branch instead of the merge commit.
```
      - uses: actions/checkout@master
        with:
          ref: ${{ github.head_ref }}
```

ericsciple:
@peter-evans instead of refs/pull/merge/\<number\> iirc refs/pull/\<number\> contains the commit (for forks and non forks). i may be wrong?

peter-evans:
@ericsciple I tested the following but it didn't work and gave the error `##[error]fatal: couldn't find remote ref refs/pull/281`

```
on: pull_request
jobs:
  pullRequest:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1
        with:
          ref: refs/pull/${{ github.event.number }}
```
[...]

PR #102, opened 2019-12-10
Fix checkout head commit of PR

mrrobot47:
`github.event.after` fails when Pull Request has just been created. At that time, `after` or `before` commit sha's do not exist. They exist during PR `synchronize` action.

Whereas `github.event.pull_request.head.sha` will work for both PR `opened` action as well as PR `synchronize` action.

Ref: #61


ericsciple:
@mrrobot47 thanks!

Issue #97, opened 2019-12-06
Document the minimum dependencies necessary to use this action with custom containers


kevinburkemeter:
It would be nice to know the minimum amount of software necessary to run this with a custom container. For example, this project requires `git` at at least v2.18. 

I'm currently trying to debug this failure:

```
/usr/bin/docker exec  186a95cc5746a0d66133af0ae7ec9e161d903825e4c12bdf02f1a2fcede1fe14 sh -c "cat /etc/*release | grep ^ID"
Running JavaScript Action with default external tool: node12
OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused "no such file or directory": unknown
```

I suspect this means I need to add Node.js to the container image but I'm not sure.

ericsciple:
@kevinburkemeter thanks. We're going to fallback to downloading the repo using the REST API if git isnt in the PATH.

ericsciple:
will be fixed by pr https://github.com/actions/checkout/pull/104

chrispat:
@kevinburkemeter we automatically inject node into the container as that is a runtime concern you should not have to deal with.

davidkarlsen:
I'm still seeing this with actions/checkout@v2 and also `master`.
I'm running on self-hosted runner, with a custom container:
```
build:
    runs-on: self-hosted
    container: maven:3-jdk-11
```

ericsciple:
Upon re-read, this looks like the same problem as https://github.com/actions/runner/issues/334

ericsciple:
Transferring to actions/runner

This looks related to minimum dependencies to run any javascript action.

Related to https://github.com/actions/runner/issues/334 ?

PR #104, opened 2019-12-10
fallback to REST API to download repo

ericsciple:
fixes https://github.com/actions/checkout/issues/97

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,7 @@ jobs:
     steps:
       # Clone this repo
       - name: Checkout
-        uses: actions/checkout@v1 # todo: switch to V2
+        uses: actions/checkout@v2-beta
 
       # Basic checkout
       - name: Basic checkout
@@ -81,3 +81,24 @@ jobs:
       - name: Verify LFS
         shell: bash
         run: __test__/verify-lfs.sh
+
+  test-job-container:
+    runs-on: ubuntu-latest
+    container: pstauffer/curl:latest
+    steps:
+      # Clone this repo
+      # todo: after v2-beta contains the latest changes, switch this to "uses: actions/checkout@v2-beta". Also switch to "alpine:latest"
+      - name: Checkout
+        run: |
+          curl --location --user token:${{ github.token }} --output checkout.tar.gz https://api.github.com/repos/actions/checkout/tarball/${{ github.sha }}
+          tar -xzf checkout.tar.gz
+          mv */* ./
+
+      # Basic checkout
+      - name: Basic checkout
+        uses: ./
+        with:
+          ref: test-data/v2/basic
+          path: basic
+      - name: Verify basic
+        run: __test__/verify-basic.sh --archive
--- a/src/git-command-manager.ts
+++ b/src/git-command-manager.ts
@@ -3,7 +3,12 @@ import * as exec from '@actions/exec'
 import * as fshelper from './fs-helper'
 import * as io from '@actions/io'
 import * as path from 'path'
+import * as retryHelper from './retry-helper'
 import {GitVersion} from './git-version'
+
+// Auth header not supported before 2.9
+// Wire protocol v2 not supported before 2.18
+export const MinimumGitVersion = new GitVersion('2.18')
 
 export interface IGitCommandManager {
   branchDelete(remote: boolean, branch: string): Promise<void>
@@ -150,22 +155,10 @@ class GitCommandManager {
       args.push(arg)
     }
 
-    let attempt = 1
-    const maxAttempts = 3
-    while (attempt <= maxAttempts) {
-      const allowAllExitCodes = attempt < maxAttempts
-      const output = await this.execGit(args, allowAllExitCodes)
-      if (output.exitCode === 0) {
-        break
-      }
-
-      const seconds = this.getRandomIntInclusive(1, 10)
-      core.warning(
-        `Git fetch failed with exit code ${output.exitCode}. Waiting ${seconds} seconds before trying again.`
-      )
-      await this.sleep(seconds * 1000)
-      attempt++
-    }
+    const that = this
+    await retryHelper.execute(async () => {
+      await that.execGit(args)
+    })
   }
 
   getWorkingDirectory(): string {
@@ -188,22 +181,10 @@ class GitCommandManager {
   async lfsFetch(ref: string): Promise<void> {
     const args = ['lfs', 'fetch', 'origin', ref]
 
-    let attempt = 1
-    const maxAttempts = 3
-    while (attempt <= maxAttempts) {
-      const allowAllExitCodes = attempt < maxAttempts
-      const output = await this.execGit(args, allowAllExitCodes)
-      if (output.exitCode === 0) {
-        break
-      }
-
-      const seconds = this.getRandomIntInclusive(1, 10)
-      core.warning(
-        `Git lfs fetch failed with exit code ${output.exitCode}. Waiting ${seconds} seconds before trying again.`
-      )
-      await this.sleep(seconds * 1000)
-      attempt++
-    }
+    const that = this
+    await retryHelper.execute(async () => {
+      await that.execGit(args)
+    })
   }
 
   async lfsInstall(): Promise<void> {
@@ -338,13 +319,9 @@ class GitCommandManager {
     }
 
     // Minimum git version
-    // Note:
-    // - Auth header not supported before 2.9
-    // - Wire protocol v2 not supported before 2.18
-    const minimumGitVersion = new GitVersion('2.18')
-    if (!gitVersion.checkMinimum(minimumGitVersion)) {
+    if (!gitVersion.checkMinimum(MinimumGitVersion)) {
       throw new Error(
-        `Minimum required git version is ${minimumGitVersion}. Your git ('${this.gitPath}') is ${gitVersion}`
+        `Minimum required git version is ${MinimumGitVersion}. Your git ('${this.gitPath}') is ${gitVersion}`
       )
     }
 
@@ -380,16 +357,6 @@ class GitCommandManager {
     const gitHttpUserAgent = `git/${gitVersion} (github-actions-checkout)`
     core.debug(`Set git useragent to: ${gitHttpUserAgent}`)
     this.gitEnv['GIT_HTTP_USER_AGENT'] = gitHttpUserAgent
-  }
-
-  private getRandomIntInclusive(minimum: number, maximum: number): number {
-    minimum = Math.floor(minimum)
-    maximum = Math.floor(maximum)
-    return Math.floor(Math.random() * (maximum - minimum + 1)) + minimum
-  }
-
-  private async sleep(milliseconds): Promise<void> {
-    return new Promise(resolve => setTimeout(resolve, milliseconds))
   }
 }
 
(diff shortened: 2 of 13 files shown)


src/git-command-manager.ts
```diff
@@ -150,22 +155,10 @@ class GitCommandManager {
       args.push(arg)
     }
```
ericsciple: Refactored most of the lines below into a retry helper. I needed to add retry logic to a new place.

src/git-command-manager.ts
```diff
@@ -188,22 +181,10 @@ class GitCommandManager {
   async lfsFetch(ref: string): Promise<void> {
     const args = ['lfs', 'fetch', 'origin', ref]
```
ericsciple: Refactored most of the lines below into a retry helper. I needed to add retry logic to a new place.

src/git-command-manager.ts
```diff
@@ -381,16 +358,6 @@ class GitCommandManager {
     core.debug(`Set git useragent to: ${gitHttpUserAgent}`)
     this.gitEnv['GIT_HTTP_USER_AGENT'] = gitHttpUserAgent
   }
-
```
ericsciple: these two functions were moved into the retry helper

src/git-command-manager.ts
```diff
@@ -338,13 +319,9 @@ class GitCommandManager {
     }
 
     // Minimum git version
```
ericsciple: This is just reaction because i needed to export `MinimumGitVersion`. The previous comments about `minimumGitVersion` moved to the new field that is exported (top of this file)

src/git-source-provider.ts
```diff
@@ -43,92 +52,89 @@ export async function getSource(settings: ISourceSettings): Promise<void> {
...
-    !(await tryPrepareExistingDirectory(
+  // Prepare existing directory, otherwise recreate
+  if (isExisting) {
+    await prepareExistingDirectory(
       git,
       settings.repositoryPath,
       repositoryUrl,
       settings.clean
-    ))
-  ) {
```
ericsciple: To simplify how this function reads, i moved the few lines below (to delete contents) into the prepareExistingDirectory function

src/git-source-provider.ts
```diff
@@ -43,92 +52,89 @@ export async function getSource(settings: ISourceSettings): Promise<void> {
...
-    // Delete the contents of the directory. Don't delete the directory itself
-    // since it may be the current working directory.
-    core.info(`Deleting the contents of '${settings.repositoryPath}'`)
-    for (const file of await fs.promises.readdir(settings.repositoryPath)) {
-      await io.rmRF(path.join(settings.repositoryPath, file))
-    }
-  }
-
```
ericsciple: The rest of these red lines were just indented. The new logic now checks `if (!git) { download using REST } else { do this existing logic }`

src/git-source-provider.ts
```diff
@@ -43,92 +46,92 @@ export async function getSource(settings: ISourceSettings): Promise<void> {
...
-  // Disable automatic garbage collection
-  if (!(await git.tryDisableAutomaticGarbageCollection())) {
-    core.warning(
-      `Unable to turn off git automatic garbage collection. The git fetch operation may trigger garbage collection and cause a delay.`
     )
   }
 
-  // Remove possible previous extraheader
-  await removeGitConfig(git, authConfigKey)
+  if (!git) {
```
ericsciple: this `if (!git) {` block is new and just calls the new helper to download the files using the REST API

src/git-source-provider.ts
```diff
@@ -43,92 +46,92 @@ export async function getSource(settings: ISourceSettings): Promise<void> {
...
+    await githubApiHelper.downloadRepository(
+      settings.accessToken,
+      settings.repositoryOwner,
+      settings.repositoryName,
+      settings.ref,
+      settings.commit,
+      settings.repositoryPath
+    )
+  } else {
+    // Save state for POST action
```
ericsciple: this save-state line was further below, i just moved it up a little and add a stateHelper abstraction

src/git-source-provider.ts
```diff
@@ -146,79 +149,110 @@ export async function cleanup(repositoryPath: string): Promise<void> {
   await removeGitConfig(git, authConfigKey)
 }
 
-async function tryPrepareExistingDirectory(
+async function getGitCommandManager(
```
ericsciple: i moved some existing logic into this function so the caller's function reads easier.

this function just instantiates the git command manager (wrapper over git executable)

src/git-source-provider.ts
```diff
@@ -146,79 +149,110 @@ export async function cleanup(repositoryPath: string): Promise<void> {
...
+): Promise<IGitCommandManager> {
+  core.info(`Working directory is '${settings.repositoryPath}'`)
+  let git = (null as unknown) as IGitCommandManager
+  try {
+    return await gitCommandManager.CreateCommandManager(
+      settings.repositoryPath,
+      settings.lfs
+    )
+  } catch (err) {
+    // Git is required for LFS
```
ericsciple: this try catch is new. As long as the user doesnt have `lfs: true`, then we can fallback to using the REST API instead

src/git-source-provider.ts
```diff
@@ -146,79 +149,110 @@ export async function cleanup(repositoryPath: string): Promise<void> {
...
+  }
+}
+
+async function prepareExistingDirectory(
   git: IGitCommandManager,
   repositoryPath: string,
   repositoryUrl: string,
   clean: boolean
-): Promise<boolean> {
+): Promise<void> {
```
ericsciple: i refactored this function a little bit

it now handles the case where `git` is null (ie not using git executable, instead using REST API)

it also performs the delete now too, if the repo folder needs to be recreated

src/retry-helper.ts
```diff
@@ -0,0 +1,58 @@
+import * as core from '@actions/core'
+
+const defaultMaxAttempts = 3
+const defaultMinSeconds = 10
+const defaultMaxSeconds = 20
+
```
ericsciple: we had two places with retry logic, and i needed to add a third.

so i went ahead and added a helper class

src/retry-helper.ts
```diff
@@ -0,0 +1,58 @@
...
+      Math.floor(Math.random() * (this.maxSeconds - this.minSeconds + 1)) +
+      this.minSeconds
+    )
+  }
+
+  private async sleep(seconds: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, seconds * 1000))
+  }
+}
+
```
ericsciple: this is the convenience function. its just a tiny wrapper over the class.

the only reason i export the class is for unit testing

src/state-helper.ts
```diff
@@ -0,0 +1,29 @@
+import * as core from '@actions/core'
+import * as coreCommand from '@actions/core/lib/command'
+
```
ericsciple: this is a simple module to make the caller code read more cleanly

this module abstracts the intra-state stuff (pretty simple logic, but reads gross in the other files, so abstracted behind this file instead)

src/github-api-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+import * as github from '@actions/github'
+import * as io from '@actions/io'
+import * as path from 'path'
+import * as retryHelper from './retry-helper'
+import * as toolCache from '@actions/tool-cache'
+import {default as uuid} from 'uuid/v4'
+import {ReposGetArchiveLinkParams} from '@octokit/rest'
+
+const IS_WINDOWS = process.platform === 'win32'
+
```
ericsciple: pretty simple function:
- Download the archive
- Write to disk
- Extract
- Move the files to the correct place (the archive contains an extra top level folder that we need to flatten)

src/github-api-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+  const archivePath = path.join(repositoryPath, `${uniqueId}.tar.gz`)
+  await fs.promises.writeFile(archivePath, archiveData)
+  archiveData = Buffer.from('') // Free memory
+
+  // Extract archive
+  core.info('Extracting the archive')
+  const extractPath = path.join(repositoryPath, uniqueId)
+  await io.mkdirP(extractPath)
+  if (IS_WINDOWS) {
+    await toolCache.extractZip(archivePath, extractPath)
```
ericsciple: todo: test e2e on win

.github/workflows/test.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     steps:
       # Clone this repo
       - name: Checkout
-        uses: actions/checkout@v1 # todo: switch to V2
+        uses: actions/checkout@v2-beta
```
ericsciple: todo: update the docs re the new functionality

src/github-api-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+  assert.ok(
+    archiveFileNames.length == 1,
+    'Expected exactly one directory inside archive'
+  )
+  const archiveVersion = archiveFileNames[0] // The top-level folder name includes the short SHA
+  core.info(`Resolved version ${archiveVersion}`)
+  const tempRepositoryPath = path.join(extractPath, archiveVersion)
+
+  // Move the files
+  for (const fileName of await fs.promises.readdir(tempRepositoryPath)) {
```
TingluoHuang: do we need to worry about windows defender take lock on files when moving?

src/github-api-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+  repo: string,
+  ref: string,
+  commit: string
+): Promise<Buffer> {
+  const octokit = new github.GitHub(accessToken)
+  const params: ReposGetArchiveLinkParams = {
+    owner: owner,
+    repo: repo,
+    archive_format: IS_WINDOWS ? 'zipball' : 'tarball',
+    ref: commit || ref
```
TingluoHuang: what's the format of `ref` does the api allowed? refs/heads/master vs master

src/github-api-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  const response = await octokit.repos.getArchiveLink(params)
+  if (response.status != 200) {
+    throw new Error(
+      `Unexpected response from GitHub API. Status: '${response.status}'`
```
TingluoHuang: you might want to dump the response body out as well, it normally contains detail error message.

src/retry-helper.ts
```diff
@@ -0,0 +1,58 @@
...
+  private maxSeconds: number
+
+  constructor(
+    maxAttempts: number = defaultMaxAttempts,
+    minSeconds: number = defaultMinSeconds,
+    maxSeconds: number = defaultMaxSeconds
+  ) {
+    this.maxAttempts = maxAttempts
+    this.minSeconds = Math.floor(minSeconds)
+    this.maxSeconds = Math.floor(maxSeconds)
```
TingluoHuang: math.ceil(maxSeconds)?

TingluoHuang: do we need to make sure max>min?

src/github-api-helper.ts
ericsciple: both work

Hamdzadprst00:
> رفع https://github.com/actions/checkout/issues/97

Hamdzadprst00:
https://github.com/actions/checkout/pull/104#issue-536012523

PR #112, opened 2019-12-13
add more scenarios

### LINKED ISSUE #45: Checkout on pull_request -> closed event
Is there a way to checkout the code of a PR when closing the PR?

We are spinning up an environment per PR and need some config from the repo to help tear it down when closing the PR.

Currently it tries to checkout the branch based off the special github pr branch

image

### LINKED ISSUE #61: Cannot use parent specifiers in the ref

### LINKED ISSUE #81: Checkout with submodules?
How do I checkout with submodules?

Before the most recent commit to master (https://github.com/actions/checkout/commit/e347bba93bdcadab0b55e4b333254f9bb40bdb0c) I could use `submodules:true` to checkout with submodules.. but it looks like yalls just removed that?

This change caused my most recent Actions checks (https://github.com/NovaMods/nova-renderer/runs/331364250) to fail to checkout

Unless I'm missing something, it seems that yalls updated the checkout action to not support submodules and that broke my Actions check

### LINKED ISSUE #91: In master/v2 can no longer git rev-parse HEAD^ to get prior commit
Currently using packtracker which relies on being able to access the prior commit via `git rev-parse HEAD^` reverting to v1.2 works.

### LINKED ISSUE #99: Master/v2 no longer creates a folder for the specified repository.
We are using `@master` of `actions/checkout` and with the help of the `repository: ImageMagick/ImageMagick-Windows` we checkout another repository. This used to create a folder called `ImageMagick-Windows` into our `%GITHUB_WORKSPACE%` folder. But due to the changes for `v2` it now clones the repository into the folder `%GITHUB_WORKSPACE%` without creating a sub directory. Is this by design or should this be fixed? Switching back to `@v1` resolved this issue.

### LINKED ISSUE #100: v2 no longer downloads tags
As of v2, tags are no longer available in the cloned repository, whereas previously tags would have been visible. I was caught out by this, as I was referencing this Action by `@master` rather than `@v1`.

Worth mentioning somewhere in the changelog?

ericsciple:
fixes #100 
fixes #99 
fixes #91 
fixes #81 
fixes #61 
fixes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,31 +73,127 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
 ```
 <!-- end usage -->
 
+# Scenarios
+
+- [Checkout a different branch](#Checkout-a-different-branch)
+- [Checkout HEAD^](#Checkout-HEAD)
+- [Checkout multiple repos (side by side)](#Checkout-multiple-repos-side-by-side)
+- [Checkout multiple repos (nested)](#Checkout-multiple-repos-nested)
+- [Checkout multiple repos (private)](#Checkout-multiple-repos-private)
+- [Checkout pull request HEAD commit instead of merge commit](#Checkout-pull-request-HEAD-commit-instead-of-merge-commit)
+- [Checkout pull request on closed event](#Checkout-pull-request-on-closed-event)
+- [Checkout submodules](#Checkout-submodules)
+- [Fetch all tags](#Fetch-all-tags)
+- [Fetch all branches](#Fetch-all-branches)
+
 ## Checkout a different branch
 
 ```yaml
 - uses: actions/checkout@v2
   with:
-    ref: some-branch
+    ref: my-branch
 ```
 
-## Checkout a different, private repository
+## Checkout HEAD^
 
 ```yaml
 - uses: actions/checkout@v2
   with:
-    repository: myAccount/myRepository
-    ref: refs/heads/master
+    fetch-depth: 2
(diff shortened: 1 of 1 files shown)


wei:
@ericsciple could you kindly comment or add a scenario that matches the behavior of v1?

ericsciple:
@wei `git fetch` (without any additional args) will fetch all branches and tags from origin

the only reason a i listed the long form was so that folks could easily modify the command to fetch a more specific subset of tags or branches (better perf for large repos)

ericsciple:
also [git fetch](https://git-scm.com/docs/git-fetch) docs are good

PR #117, opened 2019-12-19
Update README.md

imbsky:


Margen67:
That html is needed to center the image.

PR #123, opened 2019-12-24
Add link to doc for creating and using encrypted secrets

### LINKED ISSUE #75: Inform people that repository collaborators can read PATs or discourage people from using them
Inside the README, people are encouraged to use a PAT to access repository content, but this adds a risk to the repository owner's account, allowing people to read any private repository if they get access to it. As stated in the [GitHub Documentation](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets), anyone with write access can read the PAT and all private repositories a user has access to.

People should be discouraged from using PATs and should be encouraged to keep the action in their within their own repository to prevent repository collaborators from having more access to a user's account than they should.

ericsciple:
fixes #75

### DIFF
--- a/src/misc/generate-docs.ts
+++ b/src/misc/generate-docs.ts
@@ -65,8 +65,13 @@ function updateUsage(
       let segment: string = description
       if (description.length > width) {
         segment = description.substr(0, width + 1)
-        while (!segment.endsWith(' ')) {
+        while (!segment.endsWith(' ') && segment) {
           segment = segment.substr(0, segment.length - 1)
+        }
+
+        // Trimmed too much?
+        if (segment.length < width * 0.67) {
+          segment = description
         }
       } else {
         segment = description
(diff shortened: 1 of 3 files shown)


src/misc/generate-docs.ts
```diff
@@ -65,9 +65,14 @@ function updateUsage(
       let segment: string = description
       if (description.length > width) {
         segment = description.substr(0, width + 1)
-        while (!segment.endsWith(' ')) {
+        while (!segment.endsWith(' ') && segment) {
```
ericsciple: infinite loop

src/misc/generate-docs.ts
```diff
@@ -65,9 +65,14 @@ function updateUsage(
...
           segment = segment.substr(0, segment.length - 1)
         }
+
```
ericsciple: better to spill over the margin than wrap too early

PR #128, opened 2020-01-02
fix issue checking detached when git less than 2.22

### LINKED ISSUE #121: Checkout@v2 use `Git branch --show-current` which requires git 2.22.0
The v2 checkout uses `git branch --show-current` at https://github.com/actions/checkout/blob/master/src/git-command-manager.ts#L177

The `--show-current` was introduced in git [2.22.0](https://git-scm.com/docs/git-branch/2.22.0) but the checkout@v2 only requires [2.18.0](https://github.com/actions/checkout/blob/master/src/git-command-manager.ts#L11), when using checkout@v2 in a self-hosted runner, you will get an error about the `--show-current` option.
```
"H:\Program Files\Git\cmd\git.exe" branch --show-current
##[error]error: unknown option `show-current'
```

Seems like we need to bump the min required git version?

ericsciple:
fixes #121

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4887,11 +4887,9 @@ class GitCommandManager {
     }
     isDetached() {
         return __awaiter(this, void 0, void 0, function* () {
-            // Note, this implementation uses "branch --show-current" because
-            // "rev-parse --symbolic-full-name HEAD" can fail on a new repo
-            // with nothing checked out.
-            const output = yield this.execGit(['branch', '--show-current']);
-            return output.stdout.trim() === '';
+            // Note, "branch --show-current" would be simpler but isn't available until Git 2.22
+            const output = yield this.execGit(['rev-parse', '--symbolic-full-nane', 'HEAD'], true);
+            return !output.stdout.trim().startsWith('refs/heads/');
         });
     }
     lfsFetch(ref) {
--- a/dist/index.js
+++ b/dist/index.js
@@ -4888,7 +4888,7 @@ class GitCommandManager {
     isDetached() {
         return __awaiter(this, void 0, void 0, function* () {
             // Note, "branch --show-current" would be simpler but isn't available until Git 2.22
-            const output = yield this.execGit(['rev-parse', '--symbolic-full-name', 'HEAD'], true);
+            const output = yield this.execGit(['rev-parse', '--symbolic-full-name', '--verify', '--quiet', 'HEAD'], true);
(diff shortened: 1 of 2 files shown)


dist/index.js
```diff
@@ -4887,11 +4887,9 @@ class GitCommandManager {
...
     isDetached() {
         return __awaiter(this, void 0, void 0, function* () {
-            // Note, this implementation uses "branch --show-current" because
-            // "rev-parse --symbolic-full-name HEAD" can fail on a new repo
-            // with nothing checked out.
-            const output = yield this.execGit(['branch', '--show-current']);
-            return output.stdout.trim() === '';
+            // Note, "branch --show-current" would be simpler but isn't available until Git 2.22
+            const output = yield this.execGit(['rev-parse', '--symbolic-full-nane', 'HEAD'], true);
+            return !output.stdout.trim().startsWith('refs/heads/');
```
ericsciple: still testing...

ericsciple: this isnt right... i need to figure out somethign different because `fatal: ` will flag the problem matcher

dist/index.js
```diff
@@ -4799,9 +4799,11 @@ class GitCommandManager {
     branchList(remote) {
         return __awaiter(this, void 0, void 0, function* () {
             const result = [];
-            // Note, this implementation uses "rev-parse --symbolic" because the output from
+            // Note, this implementation uses "rev-parse --symbolic-full-name" because the output from
             // "branch --list" is more difficult when in a detached HEAD state.
-            const args = ['rev-parse', '--symbolic'];
+            // Note, this implementation uses "rev-parse --symbolic-full-name" because there is a bug
```
ericsciple: i ran into this bug in git 2.18 when testing. Switching to `--symbolic-full-name` and trimming `refs/heads/` and `refs/remotes/` from the output works around the issue.

dist/index.js
```diff
@@ -4799,9 +4799,11 @@ class GitCommandManager {
     branchList(remote) {
```
ericsciple: useful script for testing:

```yaml
      - run: |
          cd ..
          Invoke-WebRequest -Uri https://github.com/git-for-windows/git/releases/download/v2.18.0.windows.1/MinGit-2.18.0-64-bit.zip -OutFile mingit.zip
          mkdir mingit
          Expand-Archive -LiteralPath .\mingit.zip -DestinationPath .\mingit
          dir .\mingit
          echo "::add-path::$PWD\mingit\cmd"
```

PR #134, opened 2020-01-09
Defaults to the default branch of the repository

arnested:
The README states:

> The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event.
> Otherwise, defaults to `master`.

But it appears the default is the default branch of the repository.

Or is just that a schedule event triggers the workflow from the default branch?


ericsciple:
[scheduled](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#scheduled-events-schedule) is latest commit on the default branch

arnested:
Then what I noticed makes perfect sense :smile:

Thank you, @ericsciple!

PR #138, opened 2020-01-16
make octokit follow proxy setting.

TingluoHuang:
https://github.com/github/c2c-actions-runtime/issues/309
https://github.com/actions/checkout/issues/139


ericsciple:
i have a similar change for the toolkit (havent published the PR yet, waiting on new http-client to be published)

PR #141, opened 2020-01-21
fix ref for pr closed event when a pr is merged

### LINKED ISSUE #136: Checkout fails on pull request closed event
Hello everyone,

So, I have a `.github/workflow/docker.yml` like this:

```
name: Maybe update Docker image

on:
  pull_request:
    types: [closed]

jobs:
  maybe_update:
    runs-on: ubuntu-latest

    if: github.event.pull_request.merged == true

    steps:
    - name: Checkout
      uses: actions/checkout@v2
```

Which is pretty basic. The goal is to run this on master after a pull request is merged. It does find the lastest commit on master (be3c9f9), as seen here:

```
/usr/bin/git config --local gc.auto 0
/usr/bin/git config --local --name-only --get-regexp http\.https\:\/\/github\.com\/\.extraheader
/usr/bin/git config --local http.https://github.com/.extraheader AUTHORIZATION: basic ***
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin be3c9f9d09e54097321f92f452923a91435b63ac
```
But when it comes to checking out, it cant:

```
From https://github.com/...
 * branch            be3c9f9d09e54097321f92f452923a91435b63ac -> FETCH_HEAD
/usr/bin/git branch --list --remote origin/master
/usr/bin/git tag --list master
Removed matchers: 'checkout-git'
##[error]A branch or tag with the name 'master' could not be found
##[error]Node run failed with exit code 1
```

We also tried tying it explicitly to master, like this:

```
name: Maybe update Docker image

on:
  pull_request:
    branches: [master]
    types: [closed]    
```

But the same problem occurs. Any ideas? (:

cheers

ericsciple:
fixes #136

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -12757,6 +12757,11 @@ function getInputs() {
         if (isWorkflowRepository) {
             result.ref = github.context.ref;
             result.commit = github.context.sha;
+            // Some events have an unqualifed ref. For example when a PR is merged (pull_request closed event),
+            // the ref is unqualifed like "master" instead of "refs/heads/master".
+            if (result.commit && result.ref && !result.ref.startsWith('refs/')) {
+                result.ref = `refs/heads/${result.ref}`;
+            }
         }
         if (!result.ref && !result.commit) {
             result.ref = 'refs/heads/master';
--- a/__test__/input-helper.test.ts
+++ b/__test__/input-helper.test.ts
@@ -75,6 +75,20 @@ describe('input-helper tests', () => {
     expect(settings.repositoryPath).toBe(gitHubWorkspace)
   })
 
+  it('qualifies ref', () => {
+    let originalContext = mockGitHub.context
+    try {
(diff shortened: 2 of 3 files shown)


dist/index.js
```diff
@@ -12757,6 +12757,11 @@ function getInputs() {
         if (isWorkflowRepository) {
             result.ref = github.context.ref;
             result.commit = github.context.sha;
+            // Some events have an unqualifed ref. For example when a PR is merged (pull_request closed event),
```
ericsciple: @andymckay is this a bug on the server? Seems like github.ref would always be qualified (e.g. `refs/heads/master` not `master`). For most events it's qualified. PR closed_event (when a PR is merged) is the only case I'm aware it's not.

I'm still planning to proceed with this fix here for resiliency, but just FYI since this seems like a bug on the server.

PR #152, opened 2020-02-11
switch to spyOn for mocks

ericsciple:


__test__/input-helper.test.ts
```diff
@@ -1,47 +1,44 @@
 import * as assert from 'assert'
+import * as core from '@actions/core'
```
ericsciple: The previous code used `genMockModule` to hijack the module loader, and then `require` the module being tested during the `beforeAll`. The `require` was deferred to `beforeAll` because I had found the import order to be unreliable when esModuleInterop is set to true (recommended tsconfig setting).

Switching to `spyOn(...).mockImplementation(...)` instead, which allows everything to be imported as normal. Rather than late-bound `require` during `beforeAll`.

ericsciple:
Going ahead and merging since this is just to align with the pattern established in setup go

PR #153, opened 2020-02-13
checkout v2 adr

ericsciple:


adrs/0153-checkout-v2.md
```diff
@@ -0,0 +1,199 @@
+# ADR 0153: Checkout v2
+
+**Date**: 2019-10-21
```
ericsciple: technically multiple revisions after this date

PR #156, opened 2020-02-13
revise adr to support ssh

ericsciple:
Proposing adding SSH support. This will help for users scripting submodule checkout.

Will open a separate ADR for a first-class `submodule` input.

### DIFF
--- a/adrs/0153-checkout-v2.md
+++ b/adrs/0153-checkout-v2.md
@@ -27,13 +27,27 @@ We want to take this opportunity to make behavioral changes, from v1. This docum
       event.  Otherwise, defaults to `master`.
   token:
     description: >
-      Auth token used to fetch the repository. The token is stored in the local
-      git config, which enables your scripts to run authenticated git commands.
-      The post-job step removes the token from the git config. [Learn more about
-      creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
+      Personal access token (PAT) used to fetch the repository. The PAT is configured
+      with the local git config, which enables your scripts to run authenticated git
+      commands. The post-job step removes the PAT. [Learn more about creating and using
+      encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
     default: ${{ github.token }}
+  ssh-key:
+    description: >
+      SSH key used to fetch the repository. SSH key is configured with the local
+      git config, which enables your scripts to run authenticated git commands.
+      The post-job step removes the SSH key. [Learn more about creating and using
+      encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
+  ssh-known-hosts:
+    description: >
+      Known hosts in addition to the user and global host key database. The public
+      SSH keys for a host may be obtained using the utility `ssh-keyscan`. For example,
+      `ssh-keyscan github.com`. The public key for github.com is always implicitly added.
+  ssh-strict:
+    description: 'Whether to perform strict host key checking'
+    default: true
   persist-credentials:
-    description: 'Whether to persist the token in the git config'
+    description: 'Whether to configure the token or SSH key with the local git config'
     default: true
   path:
     description: 'Relative path under $GITHUB_WORKSPACE to place the repository'
@@ -49,6 +63,7 @@ We want to take this opportunity to make behavioral changes, from v1. This docum
 ```
 
 Note:
+- SSH support is new
 - `persist-credentials` is new
 - `path` behavior is different (refer [below](#path) for details)
 - `submodules` was removed (error if specified; add later if needed)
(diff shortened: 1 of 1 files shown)


adrs/0153-checkout-v2.md
```diff
@@ -63,19 +64,43 @@ Note:
...
   - Lines up if we add submodule support in the future. Don't need to worry about calculating relative URLs. Just works, although needs to be persisted in each submodule git config.
-  - Users opt out of persisted credentials (`persist-credentials: false`), or can script the removal themselves (`git config --unset-all http.https://github.com/.extraheader`).
+
+#### SSH key
+
+The SSH key will be written to disk under the `$RUNNER_TEMP` directory. The SSH key will
+be removed by the action's post-job hook. Additionally, RUNNER_TEMP is cleared by the
+runner between jobs.
+
+The SSH key must be written with strict file permissions. The SSH client requires the file
```
ericsciple: will the file permissions prevent scripting authenticated git commands from a docker container action?

adrs/0153-checkout-v2.md
```diff
@@ -63,19 +64,43 @@ Note:
...
+to be read/write for the user, and not accessible by others.
+
+The SSH command will be overridden for the local git config:
```
ericsciple: Since this is local to the repo config, i think users scripting submodule checkout would need to first set the env var `export GIT_SSH_COMMAND="$(git config core.sshCommand)"`

Other option is we export `GIT_SSH_COMMAND` env var, but then stomps over each other if multiple checkout with different ssh keys.

Setting both `core.sshCommand` and `GIT_SSH_COMMAND` env var doesnt make sense - env var takes precedence.

adrs/0153-checkout-v2.md
```diff
@@ -63,19 +64,43 @@ Note:
...
+
+```
+git config core.sshCommand "ssh -i path-to-ssh-key -o CheckHostIP=no"
+```
+
+[CheckHostIP](https://linux.die.net/man/5/ssh_config):
```
ericsciple: This seems like a good option to avoid maintenance when IP address changes (otherwise workflows break). Also so it just works for on-premises server.

ericsciple: I wonder if we could expose an optional input `ssh-fingerprint` to validate the fingerprint of the public key... investigating...

ericsciple: updated with a better approach

peaceiris:
Does this proposal mean that we can also do `git push` with the SSH key? If yes, we can solve this problem: ([Github action not triggering gh-pages upon push - GitHub Community Forum](https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/td-p/26869)) without setting up a personal access token or SSH key by ourself.

Now, [the GitHub Pages Actions](https://github.com/peaceiris/actions-gh-pages) provides the SSH key input option to avoid the problem. It is useful that the setting up SSH key is supported by actions/checkout. 

Thank you.

adrs/0153-checkout-v2.md
```diff
@@ -27,13 +27,27 @@ We want to take this opportunity to make behavioral changes, from v1. This docum
...
+      commands. The post-job step removes the PAT. [Learn more about creating and using
+      encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
     default: ${{ github.token }}
+  ssh-key:
+    description: >
+      SSH key used to fetch the repository. SSH key is configured with the local
+      git config, which enables your scripts to run authenticated git commands.
+      The post-job step removes the SSH key. [Learn more about creating and using
+      encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
+  ssh-known-hosts:
```
ericsciple: questions from offline review w/ cj: what if specify by ip?

adrs/0153-checkout-v2.md
```diff
@@ -63,19 +78,52 @@ Note:
...
+The SSH command will be overridden for the local git config:
+
+```
+git config core.sshCommand "ssh -i path-to-ssh-key -o StrictHostKeyChecking=yes -o CheckHostIP=no -o UserKnownHostsFile=path-to-known-hosts"
+```
+
+When the input `ssh-strict` is set to `false`, the options `CheckHostIP` and `StrictHostKeyChecking` will not be overridden.
+
+Note:
+- When `ssh-strict` is set to `true` (default), the SSH option `CheckHostIP` can safely be disabled.
```
ericsciple: questions from offline review w/ cj: will this strategy work if clone by IP? or will CheckHostIP interfere in that case

ericsciple:
@peaceiris yes the ssh key will be persisted so you can script authenticated git commands. Will be removed at the end of the job.

adrs/0153-checkout-v2.md
ericsciple: good question. Tested and it works as expected. I set the ssh url used by git to the ip address, added output of `ssh-keyscan <IP>` the known hosts input, and everything works. Without adding the known hosts input it fails - expected.

adrs/0153-checkout-v2.md
ericsciple: It appears to work, but the runner will need to add a volume mount for RUNNER_TEMP and then set the command to use `$RUNNER_TEMP/ssh-key-file` so it can be expanded at runtime.

ericsciple: @TingluoHuang @dakale fyi - pending this ADR, i'll open a PR with this change for the runner

adrs/0153-checkout-v2.md
ericsciple: I'll confirm this assumption and move this comment to the list of notes in this section - so the rationale is captured.

adrs/0153-checkout-v2.md
ericsciple: also confirmed it works for job container + container action

resolving comment and capturing in the adr in the notes for this section

ericsciple:
reviewed offline w/ @thechrisjohnson

PR #157, opened 2020-02-14
revise adr to support submodules

ericsciple:


adrs/0153-checkout-v2.md
```diff
@@ -179,6 +182,15 @@ A better solution is:
 
 Given a source file path, walk up the directories until the first `.git/config` is found. Check if it matches the self repo (`url = https://github.com/OWNER/REPO`). If not, drop the source file path.
 
+### Submodules
```
ericsciple: This section assumes `fetch-refs` and SSH key support will be added - currently in separate PR proposals.

adrs/0153-checkout-v2.md
```diff
@@ -179,6 +182,15 @@ A better solution is:
...
+
+With both PAT and SSH key support, we should be able to provide frictionless support for
+submodules scenarios: recursive, non-recursive, relative submodule paths.
+
+When fetching submodules, follow the `fetch-depth` and `fetch-refs` settings.
+
+Credentials will be persisted in the submodules local git config too.
+
```
ericsciple: Curious about retries when fetching submodules. Currently we have retry logic around fetch.

adrs/0153-checkout-v2.md
ericsciple: I think only fetch depth applies. Will yank this part about fetch refs (not implemented yet) and revisit in ADR for fetch-refs

Issue #161, opened 2020-02-22
git diff with v2


arturomf94:
As suggested [here](https://github.community/t5/GitHub-Actions/GitHub-actions-does-not-print-git-diff/td-p/43755) getting a diff works fine with v1, but not with v2. Was this overlooked? How could one get a diff with v2? 

https://github.com/actions/checkout/issues/160 seems to indicate that this can be done by fetching first. Could one get a git diff without fetching previously?

ericsciple:
V2 was optimized to download the single commit. Proposal [here](https://github.com/actions/checkout/pull/155) for additional control over which refs are fetched.

ericsciple:
Also scenarios in readme for fetching more branches/history

ericsciple:
related to https://github.com/actions/checkout/pull/155

PR #171, opened 2020-03-01
README.md: Document how to do git diff vs. master

cclauss:
#161 and many other posts say that doing `git diff` was easy in v1 but is tricky in v2.
Can we please add two items to README.md that describes how to properly do `git diff` in v2?


murrant:
Thanks, useful info.

PR #174, opened 2020-03-02
more unit tests and corresponding refactoring

ericsciple:


ericsciple:
@thboop fyi merging additional unit tests and corresponding refactoring, from the SSH PR (which you already reviewed)

PR #184, opened 2020-03-12
persist core.sshCommand for submodules

ericsciple:


__test__/git-auth-helper.test.ts
```diff
@@ -437,15 +439,16 @@ describe('git-auth-helper tests', () => {
     }
   )
 
-  const configureSubmoduleAuth_configuresTokenWhenPersistCredentialsTrueAndSshKeyNotSet =
-    'configureSubmoduleAuth configures token when persist credentials true and SSH key not set'
+  const configureSubmoduleAuth_configuresSubmodulesWhenPersistCredentialsFalseAndSshKeyNotSet =
```
ericsciple: some of the tests for configureSubmoduleAuth were redundant so i collapsed based on input combinations:
- `persist-credentials: false`, `ssh-key: ''`
- `persist-credentials: false`, `ssh-key: 'set'`
- `persist-credentials: true`, `ssh-key: ''`
- `persist-credentials: true`, `ssh-key: 'set'`

ericsciple: Each test just tests the submoduleForeach calls so it made sense to collapse.

__test__/git-auth-helper.test.ts
```diff
@@ -549,37 +552,12 @@ describe('git-auth-helper tests', () => {
...
-
-      // Act
-      await authHelper.configureSubmoduleAuth()
-
-      // Assert
-      expect(mockSubmoduleForeach).toBeCalledTimes(1)
-      expect(mockSubmoduleForeach.mock.calls[0][0] as string).toMatch(
-        /unset-all.*insteadOf/
-      )
+      expect(mockSubmoduleForeach.mock.calls[2][0]).toMatch(/core\.sshCommand/)
```
ericsciple: assert core.sshCommand was added

dist/index.js
```diff
@@ -5122,6 +5122,7 @@ class GitAuthHelper {
         this.tokenConfigKey = `http.https://${HOSTNAME}/.extraheader`;
```
ericsciple: dont review index.js (generated)

src/git-source-provider.ts
```diff
@@ -57,6 +57,16 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
     core.info(
       `To create a local Git repository instead, add Git ${gitCommandManager.MinimumGitVersion} or higher to the PATH`
     )
+    if (settings.submodules) {
```
ericsciple: fail when fallback to REST API and submodules or ssh-key is set

thboop: Shouldn't we do this before logging 'The repository will be downloaded using the GitHub REST API'

Since errors propagate up as annotations. Consider: "Git ${gitCommandManager.MinimumGitVersion} version or greater not found in path. Input 'submodules' not supported when falling back to download using the GitHub REST API"

Same for ssh key

thboop: I really like failing as an error flow here 👍

ericsciple: Updated. To make the error actionable. I switched the order a little bit from the suggestion, to lead with the error message followed by suggested action.

Issue #183, opened 2020-03-11
Unable to check out SSH submodule with the brand new ssh-key and submodule support


daveespo:
Moving this thread from #116 over here

@ericsciple  -- just tried the new ssh-key param but no joy .. (the SSH private key that I'm passing in via Secret is one that works with the ssh-agent hack I described earlier in this thread)

```
2020-03-11T20:08:31.0562996Z Current runner version: '2.165.2'
2020-03-11T20:08:31.0564725Z Prepare workflow directory
2020-03-11T20:08:31.0791957Z Prepare all required actions
2020-03-11T20:08:31.0806683Z Download action repository 'actions/checkout@master'
2020-03-11T20:08:33.0116518Z Download action repository 'sfdx-actions/setup-sfdx@v1'
2020-03-11T20:08:33.2496238Z Download action repository 'nanasess/setup-chromedriver@master'
2020-03-11T20:08:40.5559823Z Download action repository 'actions/upload-artifact@v1'
2020-03-11T20:08:41.1828303Z ##[group]Run actions/checkout@master
2020-03-11T20:08:41.1828907Z with:
2020-03-11T20:08:41.1829353Z   lfs: true
2020-03-11T20:08:41.1831532Z   ssh-key: ***
2020-03-11T20:08:41.1831924Z   submodules: true
2020-03-11T20:08:41.1832300Z   repository: <redacted>
2020-03-11T20:08:41.1832778Z   token: ***
2020-03-11T20:08:41.1833150Z   ssh-strict: true
2020-03-11T20:08:41.1833521Z   persist-credentials: true
2020-03-11T20:08:41.1833910Z   clean: true
2020-03-11T20:08:41.1834281Z   fetch-depth: 1
2020-03-11T20:08:41.1834679Z ##[endgroup]
2020-03-11T20:08:41.6235786Z Added matchers: 'checkout-git'. Problem matchers scan action output for known warning or error strings and report these inline.
2020-03-11T20:08:41.6237082Z Syncing repository: <redacted>/patronsignup
2020-03-11T20:08:41.6237958Z Working directory is '/home/runner/work/patronsignup/patronsignup'
2020-03-11T20:08:41.6310463Z [command]/usr/bin/git version
2020-03-11T20:08:41.6458784Z git version 2.25.1
2020-03-11T20:08:41.6491554Z [command]/usr/bin/git lfs version
[...]

PR #190, opened 2020-03-19
Revise ADR: Use the auth token to fetch the workflow repo

ericsciple:
Related to issue https://github.com/actions/checkout/issues/183

### DIFF
--- a/adrs/0153-checkout-v2.md
+++ b/adrs/0153-checkout-v2.md
@@ -40,9 +40,12 @@ We want to take this opportunity to make behavioral changes, from v1. This docum
     default: ${{ github.token }}
   ssh-key:
     description: >
-      SSH key used to fetch the repository. The SSH key is configured with the local
-      git config, which enables your scripts to run authenticated git commands.
-      The post-job step removes the SSH key.
+      The SSH key to use. The SSH key is configured with the local git config, which enables
+      your scripts to run authenticated git commands. The post-job step removes the SSH key.
+
+
+      When fetching the repository the workflow is running in, the auth token will be used
+      and configured instead. This enables the SSH key to be used only for submodules.
 
 
       We recommend using a service account with the least permissions necessary.


adrs/0153-checkout-v2.md
```diff
@@ -40,9 +40,12 @@ We want to take this opportunity to make behavioral changes, from v1. This docum
     default: ${{ github.token }}
   ssh-key:
```
ericsciple: I wonder whether a separate input `prefer-token` would be better than this implicit behavior.

ericsciple: Many actions take a dependency on auth being configured. So for the workflow repo, even after the initial fetch we need to leave the remote URL configured as HTTPS.

We could always add a separate input `prefer-ssh` to let users opt in to fetching the main repo using SSH. I worry a little bit about relative URLs.

Do we need to worry about LFS? I don't it affects this decision - using the token should be fine?

adrs/0153-checkout-v2.md
```diff
@@ -40,9 +40,12 @@ We want to take this opportunity to make behavioral changes, from v1. This docum
...
     description: >
-      SSH key used to fetch the repository. The SSH key is configured with the local
-      git config, which enables your scripts to run authenticated git commands.
-      The post-job step removes the SSH key.
+      The SSH key to use. The SSH key is configured with the local git config, which enables
+      your scripts to run authenticated git commands. The post-job step removes the SSH key.
+
+
+      When fetching the repository the workflow is running in, the auth token will be used
+      and configured instead. This enables the SSH key to be used only for submodules.
```
thboop: We would still persist the ssh key not the GitHub token, correct?

adrs/0153-checkout-v2.md
thboop: If we are adding extra inputs for this scenario we may want to consider having the inputs be opt in for users who need this functionality. It would be less entangled in the mainline scenario.

Its a bit verbose but we could do:
`submodules-ssh-key`
`submodules-token` 
They would default to `ssh-key` or  `token`.

This would enable us to work on any repo (not just the repo that kicked off the action). Users who don't need this scenario would ignore these inputs (and the consequences from them)

adrs/0153-checkout-v2.md
ericsciple: Technically could be the users' PAT; assumption it's the GitHub token.

adrs/0153-checkout-v2.md
thboop: > Many actions take a dependency on auth being configured. So for the workflow repo, even after the initial fetch we need to leave the remote URL configured as HTTPS.

Can we add this information to `persist-credentials`

ericsciple: I think we're saying different things. This proposal changes the behavior for the main repo only ("the repository the workflow is running in").

ericsciple: Also i think it's ok to still configure the ssh command. Remotes that use HTTPS wont use it.

adrs/0153-checkout-v2.md
ericsciple: Seems uncommon to use PAT and SSH. The scenario here is, the main repo should just work even if SSH key doesnt have permission.

adrs/0153-checkout-v2.md
ericsciple: Discussed offline. The separate inputs `submodules-ssh-key` and `submodules-token` makes more sense.

Reasons customers would want to use ssh key for the workflow repo:
- Trigger additional run on push
- Push a commit that edits the workflow file

PR #191, opened 2020-03-19
group output

ericsciple:
Collapse log output into logical groups for a nicer experience

### DIFF
--- a/src/git-directory-helper.ts
+++ b/src/git-directory-helper.ts
@@ -1,9 +1,11 @@
+import * as assert from 'assert'
 import * as core from '@actions/core'
 import * as fs from 'fs'
 import * as fsHelper from './fs-helper'
 import * as io from '@actions/io'
 import * as path from 'path'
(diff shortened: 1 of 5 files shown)


src/git-directory-helper.ts
```diff
@@ -38,6 +44,7 @@ export async function prepareExistingDirectory(
     }
 
     try {
+      core.startGroup('Removing previously created refs, to avoid conflicts')
```
thboop: I think we need to keep mark a flag to indicate we have started a group but not ended it so we can error handle correctly on this try-catch

ericsciple: Unclosed groups dont really hurt anything. Intentionally did not complicate the code to track.

Also note, can't have nested groups. Starting a new group automatically closes any open group.

Thoughts?

ericsciple: Otherwise try/finally may make sense

Thefaceofasaba (commented):
Here

Issue #176, opened 2020-03-05
Try out the new input `submodules`


ericsciple:
Collecting feedback for the new input `submodules`.

Use `actions/checkout@master`. Refer [here](https://github.com/actions/checkout/blob/master/README.md) for usage.

SSH support landing in master soon (tomorrow).

Will update the tag `v2` after collecting feedback. Likely next week.

peaceiris:
It works well :tada:

log: https://github.com/peaceiris/hugo-test-project/runs/490347719?check_suite_focus=true#step:2:135

ericsciple:
Also seeking feedback on this proposed change https://github.com/actions/checkout/pull/178 to convert SSH URLs to HTTPS when an ssh-key is not supplied.

m-hilgendorf:
What is the approach for checking out private submodules over SSH? That solution doesn't seem to work, unless I'm doing something totally off base here: 

```yaml 
    - name: checkout hack 
      run: git config --global url."https://username:${{ secrets.USER_PWD }}@github.com/".insteadOf "git@github.com:"
    - uses: actions/checkout@master 
      with: 
        submodules: recursive 
```

For all private submodules I receive: 

```
 ##[error]fatal: repository 'https://github.com/user/repo.git/' not found
```

m-hilgendorf:
@ericsciple (apologies, I moved my comment here since it's more of an issue discussion than PR review) 

> what about git config --global url."https://github.com".insteadOf "git@github.com" 

I tried that first. It only works for public submodules, the code I posted with username/password (or with a PAT) functions on a local machine. I've also tried [this](https://github.com/actions/checkout/issues/116#issuecomment-583221947). No dice, I get `error 400`. 

The only thing I have been able to get functioning is to do [this](https://github.com/actions/checkout/issues/116#issuecomment-573880976), which is non-ideal for a number of reasons.

sjackman:
I'd appreciate a configuration option to specify a list of submodules to checkout:
`git submodule update --init module-a module-b`

vadi2:
Does not work for a submodule hosted on `bitbucket.com`, see https://github.com/Mudlet/Mudlet/runs/497081182?check_suite_focus=true:

```
HEAD is now at 77edfbc Merge 3082a1eca03e73f9e3141aa33297306ff9d73a60 into 24c22e0a7878984bc105db3048db1ace9730bff1
[...]

PR #196, opened 2020-03-24
Remove submodules from readme

martinitus:
Submodules are no longer supported in checkout v2

Keep up the nice work with GH actions ;-)


martinitus:
Oh, i just realized, that while it was removed in v2, it was readded again? If so, just ignore this PR.
Sorry for the noise..

peaceiris:
We got the option again in v2.
See https://github.com/actions/checkout/issues/176

PR #201, opened 2020-03-28
Remove mention to beta for v2 in CHANGELOG and add link in README

AdrieanKhisbe:
Two very minor edits, but I found useful enough to make a PR.

This would help when searching for what were the difference between checkout v1 and v2.


AdrieanKhisbe:
Suggested edit seem to have been applied at some point.
Closing

Issue #93, opened 2019-12-04
In master/v2, cannot checkout PR base sha


mydea:
I guess this is related to #15 - I used to be able to get the base commit of a PR (to compare code coverage - here: https://github.com/mydea/ember-cli-code-coverage-action/) and check it out - which stopped working in master/v2. After some digging, I figured it was because that commit was not fetched anymore.

How would I need to configure v2 to make this work again? Thank you!

ericsciple:
Checkout v2 was optimized for the mainline case, fetch only the single commit.

If it's just the base commit that you need, you can fetch that single commit:

```
git fetch -c protocol.version=2 fetch --no-tags --progress --no-recurse-submodules --depth=1 origin ${{ github.context.pull_request.base.sha }}
```

Note, this works because `v2-beta` leaves the auth token in the git config by default, so scripting authenticated git commands just works.

Thoughts?

I also wonder whether an input to control the refspec would be good. I have some thoughts, but not fully fleshed out yet.

mydea:
Ah, nice. I'll try it later :) Maybe it would be nice to add things like this to the docs/readme, as I guess it is not an uncommon use case (to compare something for a PR)!

mansona:
I feel like this would be something to support as part of the default API for this action 🤔 there are plenty of people that will be writing Actions that are trying to do diffs between the head of a PR and master.

something like `includeBaseBranch` or something might be a reasonable addition

ericsciple:
Spitballing...

I'm thinking something like an input `refspec` which allows additional refspecs to be appended to the fetch command.

With some easy options built-in like: `branches`, `tags`, `all`, or totally custom (i.e. is a SHA or contains a `:`, multiline for multiple)

- if `branches` then the refspec `refs/heads/*:refs/remotes/origin/heads/*` is added
- if `tags` then `refs/tags/*:refs/tags/*` is added
- if `all` then `branches` + `tags`

maybe another option `base` ? only works for pull requests, i.e. `<BASE_SHA>:<BASE_REF>`

mansona:
Yea that works for me 👍 

@ericsciple how do you feel about `branches` being the default in this context (or maybe `base`)? If we do base as the default then it can just ignore if we're not on a PR

ericsciple:
@mansona I think that's fair. The goal of checkout v2 was perf improvements, so we switched to fetching a single commit by default. For a PR I don't think fetching two commits by default is bad (most git objects likely shared by the two commits anyway).

[...]

PR #213, opened 2020-04-09
Include common example for fetching PR base-ref

jasonkarns:
Since this seems to be a common scenario and is generating a lot of discussion (#93), it seems worthwhile to add it to the readme as an example.

I placed it before the other fetch examples because this use-case still only fetches minimal history (only adding one additional ref); so it seems helpful to place this above the other fetch examples which all do deeper fetches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,6 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
 
 - [Checkout a different branch](#Checkout-a-different-branch)
 - [Checkout HEAD^](#Checkout-HEAD)
-
 - [Checkout multiple repos (side by side)](#Checkout-multiple-repos-side-by-side)
 - [Checkout multiple repos (nested)](#Checkout-multiple-repos-nested)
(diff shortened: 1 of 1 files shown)


jasonkarns:
I haven't verified the table-of-contents link (I typed it by hand)

README.md
```diff
@@ -207,6 +208,15 @@ jobs:
       - uses: actions/checkout@v2
 ```
 
+## Fetch pull request base-ref in addition to merge commit (for diffing)
+
+```yml
+- uses: actions/checkout@v2
+- name: Fetch base_ref HEAD
+  run: git fetch --depth=1 origin +refs/heads/${{github.base_ref}}:refs/remotes/origin/${{github.base_ref}}
```
jasonkarns, line 216: I think this could be simplified to 
```
git fetch --depth=1 origin +${{github.base_ref}}
```
or
```
git fetch --depth=1 origin +refs/heads/${{github.base_ref}}
```

since the destination is the conventional remote ref.

ericsciple:
@jasonkarns thanks will look over this soon!

Issue #143, opened 2020-01-23
Git error after some executions


Maykonn:
The checkout action works fine (do the private repo checkout) until be executed I'm sure but something like 4 or 5 times. I mean, after some executions I'm getting the following output error of retries:

```
Run actions/checkout@v2
  with:
    repository: MyPrivate/Repo
    token: ***      # IT IS A GITHUB Personal Access Token
    ref: SRQ-644/SRQ-1099_php7_upgrade'
    path: test/test-path
    fetch-depth: 1
    persist-credentials: true
    clean: true
    lfs: false
Added matchers: 'checkout-git'. Problem matchers scan action output for known warning or error strings and report these inline.
Syncing repository: MyPrivate/Repo
Working directory is '/home/runner/work/TestActionRepo/TestActionRepo/test/test-path'
/usr/bin/git version
git version 2.24.1
/usr/bin/git init /home/runner/work/TestActionRepo/TestActionRepo/test/test-path
Initialized empty Git repository in /home/runner/work/TestActionRepo/TestActionRepo/test/test-path/.git/
/usr/bin/git remote add origin https://github.com/MyPrivate/Repo # IT IS HYPOTHETICAL PRIVATE REPO URL
/usr/bin/git config --local gc.auto 0
/usr/bin/git config --local --name-only --get-regexp http\.https\:\/\/github\.com\/\.extraheader
/usr/bin/git config --local http.https://github.com/.extraheader AUTHORIZATION: basic ***
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/SRQ-644/SRQ-1099_php7_upgrade'*:refs/remotes/origin/SRQ-644/SRQ-1099_php7_upgrade'* +refs/tags/SRQ-644/SRQ-1099_php7_upgrade'*:refs/tags/SRQ-644/SRQ-1099_php7_upgrade'*
The process '/usr/bin/git' failed with exit code 1
Waiting 17 seconds before trying again
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/SRQ-644/SRQ-1099_php7_upgrade'*:refs/remotes/origin/SRQ-644/SRQ-1099_php7_upgrade'* +refs/tags/SRQ-644/SRQ-1099_php7_upgrade'*:refs/tags/SRQ-644/SRQ-1099_php7_upgrade'*
[...]

PR #215, opened 2020-04-13
fetch verbose

ericsciple:
Increase verbosity for fetch. Currently we have cases where fetch is failing without any output. Refer to https://github.com/actions/checkout/issues/143

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -5560,7 +5560,8 @@ class GitCommandManager {
                 '--no-tags',
                 '--prune',
                 '--progress',
-                '--no-recurse-submodules'
+                '--no-recurse-submodules',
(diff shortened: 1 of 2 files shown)


dist/index.js
```diff
@@ -5560,7 +5560,8 @@ class GitCommandManager {
                 '--no-tags',
                 '--prune',
                 '--progress',
-                '--no-recurse-submodules'
+                '--no-recurse-submodules',
+                '--verbose'
```
TingluoHuang, line 5564: how much info will this print? will it help with the exit 1 case?
Should we export `GITHUB_TRACE=1` and `GIT_CURL_VERBOSE=1` when we retry for the last time?

PR #236, opened 2020-05-07
consume new @actions/github for GHES support

ericsciple:
Let the toolkit set the baseUrl (GHES vs Dotcom) when constructing the Octokit REST client

### DIFF
--- a/src/github-api-helper.ts
+++ b/src/github-api-helper.ts
@@ -6,9 +6,8 @@ import * as io from '@actions/io'
 import * as path from 'path'
 import * as retryHelper from './retry-helper'
 import * as toolCache from '@actions/tool-cache'
-import * as urlHelper from './url-helper'
 import {default as uuid} from 'uuid/v4'
-import {ReposGetArchiveLinkParams} from '@octokit/rest'
+import {Octokit} from '@octokit/rest'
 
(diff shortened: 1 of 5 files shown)


src/github-api-helper.ts
```diff
@@ -75,8 +74,8 @@ async function downloadArchive(
   ref: string,
   commit: string
 ): Promise<Buffer> {
-  const octokit = new github.GitHub(authToken, {baseUrl: urlHelper.getApiUrl()})
-  const params: ReposGetArchiveLinkParams = {
+  const octokit = new github.GitHub(authToken)
```
ericsciple, line 77: Let the toolkit set the `baseUrl`

src/github-api-helper.ts
```diff
@@ -6,9 +6,8 @@ import * as io from '@actions/io'
 import * as path from 'path'
 import * as retryHelper from './retry-helper'
 import * as toolCache from '@actions/tool-cache'
-import * as urlHelper from './url-helper'
 import {default as uuid} from 'uuid/v4'
-import {ReposGetArchiveLinkParams} from '@octokit/rest'
+import {Octokit} from '@octokit/rest'
```
ericsciple, line 10: wasnt able to import directly anymore (compat break)

PR #251, opened 2020-05-20
fix readme

ericsciple:


pllim:
I was actually looking at this one though because I was only interested in the tags, not so much branches:
```yaml
## Fetch all tags
```yaml
- uses: actions/checkout@v2
- run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
```

Seems like that section is not changed in this PR.

PR #253, opened 2020-05-20
telemetry for incorrect merge commit

ericsciple:
Adding temporary telemetry because we're sometimes receiving the wrong merge commit. Adding telemetry here, since only place that already has enough info to detect (commit message).


ericsciple:
reviewed w/ @thboop

PR #257, opened 2020-05-23
Check Git Operations availability before checking out example

crazy-max:
This is a simple example to check Git Operations availability before checking out using the [GitHub Status action](https://github.com/crazy-max/ghaction-github-status).


ericsciple:
The Actions marketplace is a good place for this action

Issue #250, opened 2020-05-18
Inconsistencies using git describe --dirty


umarcor:
While trying to use `git describe --dirty` to get a description of the repo, I found the behaviour of v2 misleading. At the same time, there are certain setups that produce a "dirty" state on Windows. To test it, I configured ~40 jobs in https://github.com/umarcor/ghdl/blob/checkout-dirty/.github/workflows/checkout.yml. Results are summarized in the following table:

| job                           | actions/checkout | unshallow | depth 0 | inherit PATH | clone in MSYS2 | pacman git | dirty? | msys dirty? |
|-------------------------------|------------------|---|---|---|---|---|----|---|
| v1 (ubuntu)                      | v1     | - | - | - | - | - | no | - |
| v1 (windows)                     | v1     | - | - | - | - | - | no | - |
| v1 (macos)                       | v1     | - | - | - | - | - | no | - |
| v2 (ubuntu)                      | v2     | - | - | - | - | - | fatal: No names found, cannot describe anything | - |
| v2 (windows)                     | v2     | - | - | - | - | - | fatal: No names found, cannot describe anything | - |
| v2 (macos)                       | v2     | - | - | - | - | - | fatal: No names found, cannot describe anything | - |
| v2-depth0 (ubuntu)               | v2     | - | x | - | - | - | fatal: No names found, cannot describe anything | - |
| v2-depth0 (windows)              | v2     | - | x | - | - | - | fatal: No names found, cannot describe anything | - |
| v2-depth0 (macos)                | v2     | - | x | - | - | - | fatal: No names found, cannot describe anything | - |
| v2-unshallow (ubuntu)            | v2     | x | - | - | - | - | no | - |
| v2-unshallow (windows)           | v2     | x | - | - | - | - | no | - |
| v2-unshallow (macos)             | v2     | x | - | - | - | - | no | - |
| v2-depth0-unshallow (ubuntu)     | v2     | x | x | - | - | - | fatal: --unshallow on a complete repository does not make sense | - |
| v2-depth0-unshallow (windows)    | v2     | x | x | - | - | - | fatal: --unshallow on a complete repository does not make sense | - |
| v2-depth0-unshallow (macos)      | v2     | x | x | - | - | - | fatal: --unshallow on a complete repository does not make sense | - |
| master (ubuntu)                  | master | - | - | - | - | - | fatal: No names found, cannot describe anything | - |
[...]

Issue #217, opened 2020-04-15
RFC: `fetch-depth: 1` and not cloning tags are dangerous defaults


sarpik:
I've had huge problems with github actions, specifically because of this action - `@actions/checkout`, and the defaults it has.

Lerna ([semantic release under the hood](https://github.com/semantic-release/semantic-release/issues/1526)) will create an incorrect version bump when used inside github actions.

This happens because by default, the https://github.com/actions/checkout will only fetch the git repository with `--depth=1` and will not fetch the tags.

In order to fix this, you need to make the following changes to your workflow:

```diff
steps:
  - uses: actions/checkout@v2
+    with:
+      # pulls all commits (needed for lerna / semantic release to correctly version)
+      fetch-depth: "0"

+ # pulls all tags (needed for lerna / semantic release to correctly version)
+ - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
```

Cheers! Hopefully I'll save someone else's time, because for me this has been a real struggle for a while.

---

I've already created FYIs in https://github.com/lerna/lerna/issues/2542 and https://github.com/semantic-release/semantic-release/issues/1526, since these are the tools I've used when I encountered the issue, but I'd also like to discuss it here.

I've previously used CircleCI and that was never an issue, but ever since I've switched to github actions, this has been secretly hiding & I didn't even notice that there were incorrect version bumps produced - I'm bringing attention to this to hopefully save some time for others, or just get the issue fixed in the first place:D

---

The fix I found was from here: https://stackoverflow.com/questions/60180630/lerna-always-lists-all-packages-ready-to-publish-when-running-workflow-of-github

and the funny thing is *how* I found it:

I've noticed that  `lerna` runs & versions incorrectly [in my github actions workflow](https://github.com/sarpik/turbo-schedule/runs/587372353?check_suite_focus=true), but once running `lerna version` locally, everything works fine.

I've checked the outputs of lerna locally and in the CI, and there was a slight difference (red - inside github actions - bad; green - locally - good):

```diff
4c4
4c4
< lerna info Assuming all packages changed
---
> lerna info Looking for changed packages since v2.1.5
8,12c8,11
<  - @turbo-schedule/client: 2.1.5 => 2.1.6
<  - @turbo-schedule/common: 2.1.5 => 2.1.6
<  - @turbo-schedule/database: 2.1.5 => 2.1.6
<  - @turbo-schedule/scraper: 2.1.5 => 2.1.6
[...]

PR #155, opened 2020-02-13
adr support fetch-refs

ericsciple:

PR #258, opened 2020-05-24
Fetch all history for all tags and branches when fetch-depth=0

### LINKED ISSUE #93: In master/v2, cannot checkout PR base sha

### LINKED ISSUE #204: Feature request: explicit fetch-depth:0 to also fetch tags
v2 is great! It's much faster and optimized for most "main" scenarios of running tests or builds on the latest commit.

However, on tag events, a very common workflow is to publish the tag as a release somewhere (homebrew, npm, github releases, etc). A major component of most release flows is generating release notes which requires knowledge of prior tags.

Using v2, I expected `fetch-depth:0` to make this possible.

Problematic run: https://github.com/nodenv/nodenv/runs/549286319?check_suite_focus=true

Error:
```
Run script/release-notes | tee relnotes.txt
fatal: No tags can describe '443076f046e3db1820d725f6d82947c55dcbdc8f'.
Try --always, or create some tags.
```

workflow:

```yml
    steps:
      - uses: actions/checkout@v2
        with: { fetch-depth: 0 }
      - run: script/release-notes | tee relnotes.txt
```

My hope is that when fetch-depth:0 is supplied (which implies we want _all history_) that tags will be fetched as well. I believe this would keep the non-tag "main" scenario fast as desired; while also making the most common tagging scenario more intuitive.

related to #100

### LINKED ISSUE #214: Support multiple branches being checked out through config
I've been working on Actions CI that leverages both `actions/checkout` and Jest for tests.

I am specifically using Jest's [`changedSince`](https://jestjs.io/docs/en/cli.html#--changedsince) flag, which compares the current branch with the passed branch and limits the tests it is running to only those that are associated with the files changed. This is particularly useful for only running the relevant tests when compared against `master` in a PR.

It would significantly reduce the necessary work for users if there was a way to include multiple branches in the run through a simple YAML property. At present, I need to do this:

```yaml
    steps:
    - uses: actions/checkout@v2
    - run: |
        git fetch --no-tags --depth=1 origin master
        git checkout -b master
        git checkout ${{ github.event.pull_request.head.sha }}
```

something like the following would be amazingly useful, where `checkout` is an array of branches to checkout on every run _while still keeping the PR's branch as the currently checked out branch_:

```
    steps:
    - uses: actions/checkout@v2
       checkout: ['master']
```

`checkout` may not be the most appropriate name, but the idea is more of what I'm going for here ❤️

### LINKED ISSUE #239: recursive checkout action not work as exptected.
I tried to use
```yaml
    - uses: actions/checkout@v2
      with:
        submodules: 'recursive'
```
or
```yaml
    - uses: actions/checkout@v2
      with:
        submodules: 'recursive'
        fetch-depth: '0'
```
But during the latter build process, it always failed with something error message like
```
Error: Command failed: git describe --tags
fatal: No names found, cannot describe anything.
```
see https://github.com/liudonghua123/dpo-voyager/runs/662309034?check_suite_focus=true.

And in the end I need to rewrite the checkout steps like the following to make the build process success.
```yaml
    - uses: actions/checkout@v2
      with:
        submodules: 'recursive'
    - run: git fetch --prune --unshallow
```

ref: https://stackoverflow.com/questions/4916492/git-describe-fails-with-fatal-no-names-found-cannot-describe-anything

Is there any options to make git work as `--prune --unshallow`?

### LINKED ISSUE #240: Fetch tags by default
checkout@v1 used to pull all tags by default. It seems that checkout@v2 passes --no-tags flag and does not pull the tags by default. Can you bring back the old behavior because we use those tags to tag docker images etc. This seems like an unusual default.

ericsciple:
Fetch all history for all tags and branches when `fetch-depth: 0`

Fixes #240
Fixes #239
Fixes #214
Fixes #204
Fixes #93
Related #250
Related #217

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 This action checks-out your repository under `$GITHUB_WORKSPACE`, so your workflow can access it.
 
-Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set `fetch-depth` to fetch more history. Refer [here](https://help.github.com/en/articles/events-that-trigger-workflows) to learn which commit `$GITHUB_SHA` points to for different events.
+Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set `fetch-depth: 0` to fetch all history for all branches and tags. Refer [here](https://help.github.com/en/articles/events-that-trigger-workflows) to learn which commit `$GITHUB_SHA` points to for different events.
 
 The auth token is persisted in the local git config. This enables your scripts to run authenticated git commands. The token is removed during post-job cleanup. Set `persist-credentials: false` to opt-out.
 
@@ -110,6 +110,7 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
 
 # Scenarios
 
+- [Fetch all history for all tags and branches](#Fetch-all-history-for-all-tags-and-branches)
 - [Checkout a different branch](#Checkout-a-different-branch)
 - [Checkout HEAD^](#Checkout-HEAD)
 - [Checkout multiple repos (side by side)](#Checkout-multiple-repos-side-by-side)
@@ -117,9 +118,14 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
 - [Checkout multiple repos (private)](#Checkout-multiple-repos-private)
 - [Checkout pull request HEAD commit instead of merge commit](#Checkout-pull-request-HEAD-commit-instead-of-merge-commit)
 - [Checkout pull request on closed event](#Checkout-pull-request-on-closed-event)
-- [Fetch all tags](#Fetch-all-tags)
-- [Fetch all branches](#Fetch-all-branches)
-- [Fetch all history for all tags and branches](#Fetch-all-history-for-all-tags-and-branches)
+
+## Fetch all history for all tags and branches
+
+```yaml
+- uses: actions/checkout@v2
+  with:
+    fetch-depth: 0
+```
 
 ## Checkout a different branch
 
@@ -205,29 +211,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-```
-
-## Fetch all tags
-
-```yaml
-- uses: actions/checkout@v2
-- run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
-```
-
-## Fetch all branches
-
-```yaml
-- uses: actions/checkout@v2
-- run: |
-    git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
-```
-
-## Fetch all history for all tags and branches
-
-```yaml
-- uses: actions/checkout@v2
-- run: |
-    git fetch --prune --unshallow --tags
 ```
 
 # License
--- a/dist/index.js
+++ b/dist/index.js
@@ -3383,6 +3383,7 @@ Object.defineProperty(exports, "__esModule", { value: true });
 const url_1 = __webpack_require__(835);
 const core = __importStar(__webpack_require__(470));
 const github = __importStar(__webpack_require__(469));
+exports.tagsRefSpec = '+refs/tags/*:refs/tags/*';
 function getCheckoutInfo(git, ref, commit) {
     return __awaiter(this, void 0, void 0, function* () {
         if (!git) {
@@ -3429,6 +3430,15 @@ function getCheckoutInfo(git, ref, commit) {
     });
 }
 exports.getCheckoutInfo = getCheckoutInfo;
+function getRefSpecForAllHistory(ref, commit) {
+    const result = ['+refs/heads/*:refs/remotes/origin/*', exports.tagsRefSpec];
+    if (ref && ref.toUpperCase().startsWith('REFS/PULL/')) {
+        const branch = ref.substring('refs/pull/'.length);
+        result.push(`+${commit || ref}:refs/remotes/pull/${branch}`);
+    }
+    return result;
+}
+exports.getRefSpecForAllHistory = getRefSpecForAllHistory;
 function getRefSpec(ref, commit) {
     if (!ref && !commit) {
         throw new Error('Args ref and commit cannot both be empty');
@@ -3478,6 +3488,50 @@ function getRefSpec(ref, commit) {
     }
 }
 exports.getRefSpec = getRefSpec;
+/**
+ * Tests whether the initial fetch created the ref at the expected commit
+ */
+function testRef(git, ref, commit) {
+    return __awaiter(this, void 0, void 0, function* () {
+        if (!git) {
+            throw new Error('Arg git cannot be empty');
+        }
+        if (!ref && !commit) {
+            throw new Error('Args ref and commit cannot both be empty');
+        }
+        // No SHA? Nothing to test
+        if (!commit) {
+            return true;
+        }
+        // SHA only?
+        else if (!ref) {
+            return yield git.shaExists(commit);
+        }
+        const upperRef = ref.toUpperCase();
+        // refs/heads/
+        if (upperRef.startsWith('REFS/HEADS/')) {
+            const branch = ref.substring('refs/heads/'.length);
+            return ((yield git.branchExists(true, `origin/${branch}`)) &&
+                commit === (yield git.revParse(`refs/remotes/origin/${branch}`)));
+        }
+        // refs/pull/
+        else if (upperRef.startsWith('REFS/PULL/')) {
+            // Assume matches because fetched using the commit
+            return true;
+        }
+        // refs/tags/
+        else if (upperRef.startsWith('REFS/TAGS/')) {
+            const tagName = ref.substring('refs/tags/'.length);
+            return ((yield git.tagExists(tagName)) && commit === (yield git.revParse(ref)));
+        }
+        // Unexpected
+        else {
+            core.debug(`Unexpected ref format '${ref}' when testing ref info`);
+            return true;
+        }
+    });
+}
+exports.testRef = testRef;
 function checkCommitInfo(token, commitInfo, repositoryOwner, repositoryName, ref, commit) {
     return __awaiter(this, void 0, void 0, function* () {
         try {
@@ -5634,6 +5688,7 @@ const exec = __importStar(__webpack_require__(986));
 const fshelper = __importStar(__webpack_require__(618));
 const io = __importStar(__webpack_require__(1));
 const path = __importStar(__webpack_require__(622));
+const refHelper = __importStar(__webpack_require__(227));
 const regexpHelper = __importStar(__webpack_require__(528));
 const retryHelper = __importStar(__webpack_require__(587));
 const git_version_1 = __webpack_require__(559);
@@ -5749,18 +5804,14 @@ class GitCommandManager {
             return output.exitCode === 0;
         });
     }
-    fetch(fetchDepth, refSpec) {
+    fetch(refSpec, fetchDepth) {
         return __awaiter(this, void 0, void 0, function* () {
-            const args = [
-                '-c',
-                'protocol.version=2',
-                'fetch',
-                '--no-tags',
-                '--prune',
-                '--progress',
-                '--no-recurse-submodules'
-            ];
-            if (fetchDepth > 0) {
+            const args = ['-c', 'protocol.version=2', 'fetch'];
+            if (!refSpec.some(x => x === refHelper.tagsRefSpec)) {
+                args.push('--no-tags');
+            }
+            args.push('--prune', '--progress', '--no-recurse-submodules');
+            if (fetchDepth && fetchDepth > 0) {
                 args.push(`--depth=${fetchDepth}`);
             }
             else if (fshelper.fileExistsSync(path.join(this.workingDirectory, '.git', 'shallow'))) {
@@ -5819,8 +5870,27 @@ class GitCommandManager {
     removeEnvironmentVariable(name) {
         delete this.gitEnv[name];
     }
+    /**
+     * Resolves a ref to a SHA. For a branch or lightweight tag, the commit SHA is returned.
+     * For an annotated tag, the tag SHA is returned.
+     * @param {string} ref  For example: 'refs/heads/master' or '/refs/tags/v1'
+     * @returns {Promise<string>}
+     */
+    revParse(ref) {
+        return __awaiter(this, void 0, void 0, function* () {
+            const output = yield this.execGit(['rev-parse', ref]);
+            return output.stdout.trim();
+        });
+    }
     setEnvironmentVariable(name, value) {
         this.gitEnv[name] = value;
+    }
+    shaExists(sha) {
+        return __awaiter(this, void 0, void 0, function* () {
+            const args = ['rev-parse', '--verify', '--quiet', `${sha}^{object}`];
+            const output = yield this.execGit(args, true);
+            return output.exitCode === 0;
+        });
     }
     submoduleForeach(command, recursive) {
         return __awaiter(this, void 0, void 0, function* () {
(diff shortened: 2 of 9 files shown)


src/git-source-provider.ts
```diff
@@ -109,8 +110,17 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
 
     // Fetch
     core.startGroup('Fetching the repository')
-    const refSpec = refHelper.getRefSpec(settings.ref, settings.commit)
-    await git.fetch(settings.fetchDepth, refSpec)
+    if (settings.fetchDepth <= 0) {
+      let refSpec = refHelper.getRefSpecForAllHistory(settings.ref)
+      await git.fetch(refSpec)
+      if (settings.commit && !(await git.shaExists(settings.commit))) {
```
ericsciple, line 116: tested the shaExists fallback E2E using a force pushed branch, and using a force pushed PR

src/git-directory-helper.ts
```diff
@@ -56,10 +56,24 @@ export async function prepareExistingDirectory(
         await git.branchDelete(false, branch)
       }
 
-      // Remove all refs/remotes/origin/* to avoid conflicts
-      branches = await git.branchList(true)
-      for (const branch of branches) {
-        await git.branchDelete(true, branch)
+      // Remove any conflicting refs/remotes/origin/*
```
ericsciple, line 59: Detects two types of conflicts:

Example 1: Consider `ref` is `refs/heads/foo` and an existing ref `refs/remotes/origin/foo/bar` exists

Example 2: Consider `ref` is `refs/heads/foo/bar` and an existing ref `refs/remotes/origin/foo` exists

ericsciple: todo: add examples to comments in the code

src/git-command-manager.ts
```diff
@@ -242,6 +241,12 @@ class GitCommandManager {
     this.gitEnv[name] = value
   }
 
+  async shaExists(sha: string): Promise<boolean> {
+    const args = ['rev-parse', '--verify', '--quiet', `${sha}^{object}`]
```
ericsciple, line 245: rev-parse magic :)

`^{object}` means the object can be any type

ericsciple: i love git :)

it's a great command line tool that lets you do what you need to do

src/git-command-manager.ts
```diff
@@ -164,17 +166,14 @@ class GitCommandManager {
...
-      '--no-tags',
-      '--prune',
-      '--progress',
-      '--no-recurse-submodules'
-    ]
-    if (fetchDepth > 0) {
+  async fetch(refSpec: string[], fetchDepth?: number): Promise<void> {
+    const args = ['-c', 'protocol.version=2', 'fetch']
+    if (!refSpec.some(x => x === refHelper.tagsRefSpec)) {
+      args.push('--no-tags')
```
ericsciple, line 172: not adding `--no-tags` when `+refs/tags/*:refs/tags/*` is part of the refspec... i think it would technically work the same, but would just look weird

README.md
```diff
@@ -207,29 +213,6 @@ jobs:
       - uses: actions/checkout@v2
 ```
 
-## Fetch all tags
```
ericsciple, line 210: bye bye complicated stuff

README.md
```diff
@@ -110,16 +110,22 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
...
 - [Checkout multiple repos (side by side)](#Checkout-multiple-repos-side-by-side)
 - [Checkout multiple repos (nested)](#Checkout-multiple-repos-nested)
 - [Checkout multiple repos (private)](#Checkout-multiple-repos-private)
 - [Checkout pull request HEAD commit instead of merge commit](#Checkout-pull-request-HEAD-commit-instead-of-merge-commit)
 - [Checkout pull request on closed event](#Checkout-pull-request-on-closed-event)
-- [Fetch all tags](#Fetch-all-tags)
-- [Fetch all branches](#Fetch-all-branches)
-- [Fetch all history for all tags and branches](#Fetch-all-history-for-all-tags-and-branches)
+
+## Fetch all history for all tags and branches
```
ericsciple, line 122: @chrispat fyi

ericsciple: [this comment](https://github.com/actions/checkout/issues/217#issuecomment-621967093) sums up the motivation nicely:

> Seems pretty logical and reasonable to me that if I set fetch-depth: 0 then I clearly want everything and history matters to me and that you should not add the --no-tags parameter.

dist/index.js
```diff
@@ -3429,6 +3430,18 @@ function getCheckoutInfo(git, ref, commit) {
     });
 }
 exports.getCheckoutInfo = getCheckoutInfo;
+function getRefSpecForAllHistory(ref) {
+    const result = ['+refs/heads/*:refs/remotes/origin/*', exports.tagsRefSpec];
```
ericsciple, line 3434: branches and tags

dist/index.js
```diff
@@ -3429,6 +3430,18 @@ function getCheckoutInfo(git, ref, commit) {
...
+    if (ref) {
+        const upperRef = (ref || '').toUpperCase();
+        if (ref.toUpperCase().startsWith('REFS/PULL/')) {
```
ericsciple, line 3437: and conditionally the pull request branch

src/ref-helper.ts
```diff
@@ -60,6 +62,19 @@ export async function getCheckoutInfo(
   return result
 }
 
+export function getRefSpecForAllHistory(ref: string): string[] {
+  const result = ['+refs/heads/*:refs/remotes/origin/*', tagsRefSpec]
```
ericsciple, line 66: branches and tags

src/ref-helper.ts
```diff
@@ -60,6 +62,19 @@ export async function getCheckoutInfo(
...
+  if (ref) {
+    const upperRef = (ref || '').toUpperCase()
+    if (ref.toUpperCase().startsWith('REFS/PULL/')) {
```
ericsciple, line 69: and conditionally the pull request branch

src/git-source-provider.ts
```diff
@@ -109,8 +110,17 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
 
     // Fetch
     core.startGroup('Fetching the repository')
-    const refSpec = refHelper.getRefSpec(settings.ref, settings.commit)
-    await git.fetch(settings.fetchDepth, refSpec)
+    if (settings.fetchDepth <= 0) {
+      let refSpec = refHelper.getRefSpecForAllHistory(settings.ref)
+      await git.fetch(refSpec)
```
ericsciple, line 115: todo: test scenario where a new commit was pushed (not force push). Going to need some additional changes to this PR to checkout the correct commit

src/git-source-provider.ts
ericsciple: todo: test scenario where a new commit was pushed (not force push). Going to need some additional changes to this PR to checkout the correct commit

ericsciple: ok i think instead of `shaExists`, i need to test whether the remote ref matches the expected SHA. If not then override using fetch. That should fix all scenarios (force push and non force push)

ericsciple: might still need `shaExists` if checking out a specific commit (no ref)

src/git-source-provider.ts
```diff
@@ -109,8 +110,20 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
     core.startGroup('Fetching the repository')
-    const refSpec = refHelper.getRefSpec(settings.ref, settings.commit)
-    await git.fetch(settings.fetchDepth, refSpec)
+    if (settings.fetchDepth <= 0) {
+      let refSpec = refHelper.getRefSpecForAllHistory(
+        settings.ref,
+        settings.commit
+      )
+      await git.fetch(refSpec)
+      if (!(await refHelper.testRef(git, settings.ref, settings.commit))) {
```
ericsciple, line 119: When all history is fetched (all branches and all tags), the ref we're interested in may have moved to a different commit (push or force push)

Test whether the ref points to the expected commit. If not, then fetch again with a targeted refspec.

I'll add some comments in the code here.

src/ref-helper.ts
```diff
@@ -111,6 +126,61 @@ export function getRefSpec(ref: string, commit: string): string[] {
...
+  else if (upperRef.startsWith('REFS/PULL/')) {
+    // Assume matches because fetched using the commit
+    return true
+  }
+  // refs/tags/
+  else if (upperRef.startsWith('REFS/TAGS/')) {
+    const tagName = ref.substring('refs/tags/'.length)
+    return (
+      (await git.tagExists(tagName)) &&
+      commit === (await git.getCommitSha(commit))
```
ericsciple, line 174: i might switch to revparse instead. I think technically the `commit` parameter could be a tag SHA (not a commit sha). Either way will work, but switching to revparse more technically accurate and eliminate extra fetch for annotated tags.

dist/index.js
```diff
@@ -7494,10 +7580,24 @@ function prepareExistingDirectory(git, repositoryPath, repositoryUrl, clean) {
...
                 }
-                // Remove all refs/remotes/origin/* to avoid conflicts
-                branches = yield git.branchList(true);
-                for (const branch of branches) {
-                    yield git.branchDelete(true, branch);
+                // Remove any conflicting refs/remotes/origin/*
+                // Example 1: Consider ref is refs/heads/foo and previously fetched refs/remotes/origin/foo/bar
+                // Example 2: Consider ref is refs/heads/foo/bar and previously fetched refs/remotes/origin/foo
+                if (ref) {
+                    ref = ref.startsWith('refs/') ? ref : `refs/heads/${ref}`;
```
thboop, line 7587: Where are we handling conflicts for tags (in the case of a force push update?)

src/git-directory-helper.ts
```diff
@@ -56,10 +56,26 @@ export async function prepareExistingDirectory(
...
+      // Example 1: Consider ref is refs/heads/foo and previously fetched refs/remotes/origin/foo/bar
+      // Example 2: Consider ref is refs/heads/foo/bar and previously fetched refs/remotes/origin/foo
+      if (ref) {
+        ref = ref.startsWith('refs/') ? ref : `refs/heads/${ref}`
```
thboop, line 63: Should we do the same thing with tags? In case a force push has been made to a tag, a pull wouldn't work.

ericsciple: Note, this PR doesn't change that behavior but I did think about tags too.

Technically could happen for tags, however unlikely because tags typically aren't pathy and also commonly never move. Further evidence: we don't handle today and it hasn't been a problem. Same during azp, never saw issue for tag.

If it were to happen here is what the error looks like:

```
$ git tag "pathy/tag"
$ git tag pathy
fatal: cannot lock ref 'refs/tags/pathy': 'refs/tags/pathy/tag' exists; cannot create 'refs/tags/pathy'
```

Delete the repo and reclone is one way to fix. Otherwise when fetching all history, will be deleted because of `git fetch origin --prune '+refs/tags/*:refs/tags/*'`

src/git-source-provider.ts
```diff
@@ -109,8 +110,24 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
+      // Fetch all branches and tags
+      let refSpec = refHelper.getRefSpecForAllHistory(
+        settings.ref,
+        settings.commit
+      )
+      await git.fetch(refSpec)
+
+      // When all history is fetched, the ref we're interested in may have moved to a different
+      // commit (push or force push). If so, fetch again with a targeted refspec.
+      if (!(await refHelper.testRef(git, settings.ref, settings.commit))) {
```
ericsciple, line 123: print to console?

src/ref-helper.ts
```diff
@@ -60,6 +62,19 @@ export async function getCheckoutInfo(
   return result
 }
 
+export function getRefSpecForAllHistory(ref: string, commit: string): string[] {
+  const result = ['+refs/heads/*:refs/remotes/origin/*', tagsRefSpec]
+  if (ref) {
+    const upperRef = (ref || '').toUpperCase()
```
TingluoHuang, line 68: extra `upperRef`

src/ref-helper.ts
```diff
@@ -111,6 +126,60 @@ export function getRefSpec(ref: string, commit: string): string[] {
...
+  // refs/tags/
+  else if (upperRef.startsWith('REFS/TAGS/')) {
+    const tagName = ref.substring('refs/tags/'.length)
+    return (
+      (await git.tagExists(tagName)) && commit === (await git.revParse(ref))
+    )
+  }
+  // Unexpected
+  else {
+    core.debug(`Unexpected ref format '${ref}' when testing ref info`)
```
TingluoHuang, line 178: do we need to error on unexpected?

src/git-command-manager.ts
```diff
@@ -164,17 +167,14 @@ class GitCommandManager {
...
-      'fetch',
-      '--no-tags',
-      '--prune',
-      '--progress',
-      '--no-recurse-submodules'
-    ]
-    if (fetchDepth > 0) {
+  async fetch(refSpec: string[], fetchDepth?: number): Promise<void> {
+    const args = ['-c', 'protocol.version=2', 'fetch']
+    if (!refSpec.some(x => x === refHelper.tagsRefSpec)) {
```
TingluoHuang, line 172: do we need to worry about casing? `Refs/Tags/*` vs `refs/tags/*`

src/git-command-manager.ts
```diff
@@ -164,17 +167,14 @@ class GitCommandManager {
...
+      args.push('--no-tags')
+    }
+
+    args.push('--prune', '--progress', '--no-recurse-submodules')
+    if (fetchDepth) {
```
TingluoHuang, line 177: will `--depth=0` or `--depth=-1` hurt anything?

ericsciple:
Note to self, tested E2E:
- Branch move forward before checkout
- Branch deleted before checkout
- PR branch move forward before checkout
- Tag move before checkout
- Tag deleted before checkout
- Conflicts for branch name (both ways)

src/git-command-manager.ts
ericsciple: not worried in this case, just used to avoid `--no-tags` flag. I think would be safe to add `--no-tags` but would just look weird when tags refspec supplied. Also, customer doesn't control the refspec, it's generated from refHelper which also exports the const `tagsRefSpec`.

src/ref-helper.ts
ericsciple: will fix

src/ref-helper.ts
ericsciple: if i had confidence from telemetry i would error :)

src/git-command-manager.ts
ericsciple: Technically not a problem based on caller. However will fix to handle (better)

marcellodesales:
@ericsciple I still see the following:

```yaml
jobs:
  test:
    name: Run Test Cases
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          # https://github.com/actions/checkout/pull/258 needs to fetch all tags so that Makefile can make the correct version
          fetch-depth: 0
```

* I'm seeing the following not loading the tags... it says `--no-tags` :( and --depth=1... Is this because of my comment above?

```
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules 
  --depth=1 origin +e977ace6752fbf2dffc9bbdead153339b9b1896f:refs/remotes/origin/feature/ci-cd-cli-in-docker-image
```

GMNGeoffrey:
Ok so what can I do now if I want to fetch the history of a single branch and *not* the entire history of the entire repo? Overloading fetch-depth means that this is no longer possible (as opposed to the proposal in #155 to add fetch-refs) :-/

I see this PR originated from a number of user complaints, so I guess you're never going to be able to support all these different scenarios. Maybe what I really want is just an action to set up the auth stuff and then just use normal git commands to avoid weird changes under the hood.

chrispat:
@GMNGeoffrey that is essentially what this action does.  If you use it in the default configuration the auth should be persisted so you can just run additional Git commands.

GMNGeoffrey:
Yeah but it's *called* the checkout command and it still does checkout :-D And if I use it in its default configuration I end up with a weird grafted HEAD and have to `--unshallow`

PR #278, opened 2020-06-15
determine default branch

### LINKED ISSUE #210: support renamed default branches
the `ref:` input parameter seems to be hardcoded falling back to `master`.

https://github.com/actions/checkout/blob/master/README.md#L43-L46

since github support renaming the default branch, IMO the actions/checkout action should take this into account and therefore use the 'default-branch' as a fallback instead of always `master`

### LINKED ISSUE #221: could not read Username for 'https://github.com': terminal prompts disabled
Url to log https://github.com/trustwallet/assets/pull/1898/checks?check_run_id=597005450
```
name: Formatter
on:
  push:
    branches:
      - '*'
      - '!master'
  pull_request:
    branches:
      - 'master'
jobs:
  format:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          ref: ${{ github.head_ref }}
          token: ${{ secrets.GITHUB_API_TOKEN }}
```

ericsciple:
Fixes #210
Fixes #221

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -8,7 +8,7 @@ inputs:
     description: >
       The branch, tag or SHA to checkout. When checking out the repository that
       triggered a workflow, this defaults to the reference or SHA for that
-      event.  Otherwise, defaults to `master`.
+      event.  Otherwise, uses the default branch.
   token:
     description: >
       Personal access token (PAT) used to fetch the repository. The PAT is configured
--- a/dist/index.js
+++ b/dist/index.js
@@ -6114,6 +6114,12 @@ function getSource(settings) {
         // Repository URL
         core.info(`Syncing repository: ${settings.repositoryOwner}/${settings.repositoryName}`);
         const repositoryUrl = urlHelper.getFetchUrl(settings);
+        // Determine the default branch
+        if (!settings.ref && !settings.commit) {
+            core.startGroup('Determining the default branch');
+            settings.ref = yield githubApiHelper.getDefaultBranch(settings.authToken, settings.repositoryOwner, settings.repositoryName);
+            core.endGroup();
+        }
         // Remove conflicting file path
         if (fsHelper.fileExistsSync(settings.repositoryPath)) {
             yield io.rmRF(settings.repositoryPath);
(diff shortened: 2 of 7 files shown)


src/input-helper.ts
```diff
@@ -110,7 +106,7 @@ export function getInputs(): IGitSourceSettings {
   core.debug(`recursive submodules = ${result.nestedSubmodules}`)
 
   // Auth token
-  result.authToken = core.getInput('token')
+  result.authToken = core.getInput('token', {required: true})
```
joshmgross, line 109: Isn't this a breaking change? Does the action need a token if checking out a public repo?

If we're making this required, we should update the action.yml file with `required: true` for this input as well.

ericsciple: Yes today it already needs a token but fails with a bad error #221

ericsciple: Generally it doesnt need to be specified, because the default value works but sometimes folks have set it to `''` or due to fork PR a secret doesnt get mapped in

joshmgross: If it's empty, can I still checkout from a public repo?

ericsciple: No today it fails like this:

```
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +559278e5d4074a5495dceca4d32edaa1acc01724:refs/remotes/origin/master
##[error]fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

action.yml
```diff
@@ -8,7 +8,7 @@ inputs:
     description: >
       The branch, tag or SHA to checkout. When checking out the repository that
       triggered a workflow, this defaults to the reference or SHA for that
-      event.  Otherwise, defaults to `master`.
+      event.  Otherwise, uses the default branch.
   token:
```
joshmgross, line 12: Could we update this to have `required: true`?

portlek:
Fixes #279 and #280 too

action.yml
ericsciple: It's not required that the user enter a value. It has a default.

src/git-source-provider.ts
```diff
@@ -19,6 +19,17 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
   )
   const repositoryUrl = urlHelper.getFetchUrl(settings)
 
+  // Determine the default branch
+  if (!settings.ref && !settings.commit) {
+    core.startGroup('Determine the default branch')
```
thboop, line 24: Super nit:

All of our other groups names using the present progressive verb tense `Initializing the repository' `Initializing the repository'` `Getting Git version info`

Can we update this to ` Determining the default branch`

gr2m:
Thanks for implementing this! 🎉

dist/index.js
```diff
@@ -9569,6 +9575,31 @@ function downloadRepository(authToken, owner, repo, ref, commit, repositoryPath)
...
 exports.downloadRepository = downloadRepository;
+/**
+ * Looks up the default branch name
+ */
+function getDefaultBranch(authToken, owner, repo) {
+    return __awaiter(this, void 0, void 0, function* () {
+        return yield retryHelper.execute(() => __awaiter(this, void 0, void 0, function* () {
+            core.info('Retrieving the default branch name');
+            const octokit = new github.GitHub(authToken);
+            const response = yield octokit.repos.get({ owner, repo });
```
felixarntz, line 9586: @ericsciple We are using this action on our repository's `.wiki` repo, and our respective action has been failing recently as the API returns "Not Found" for those. Could this be an oversight here?

ericsciple: Thanks for the heads up! And sorry for the disruption :(

I will look into the scenario shortly...

felixarntz: Thanks! Fortunately was an easy fix on our end :)

Is it even possible to change the wiki default branch name? If not, maybe for those repositories the old hard-coded "logic" could be maintained for now.

Issue #283, opened 2020-06-17
Latest v2 broke all of my private github actions


derek-burdick:
After the release this morning, all of my github actions using checkout to pull a private action started failing.

```
      - uses: actions/checkout@v2
        with:
          repository: privateorg/actions
          path: .privateorg-actions
          ssh-key: ${{ secrets.GITSSHKEY }}
```

I would like to request that you revert the change that is now breaking hundreds of my private workflows.

drwoods:
Same here.  We had to update several repos (with many more to come) to use the token instead -
```
- uses: actions/checkout@v2
        with:
          repository: privateorg/actions
          path: .privateorg-actions
          token: ${{ secrets.TOKEN }}
```
The v2.3.0 release needs to be reverted and released as a breaking v3 release if token needs to be a required/valid parameter.

mizumotok:
I got the same problem.
"Determining the default branch" fails using ssh-key.
```
Determining the default branch
  Retrieving the default branch name
  Not Found
  Waiting 12 seconds before trying again
  Retrieving the default branch name
  Not Found
  Waiting 12 seconds before trying again
  Retrieving the default branch name
  ##[error]Not Found
```

I guess checking the default branch before setting up auth.

https://github.com/actions/checkout/blob/v2.3.0/src/git-source-provider.ts#L22
```typescript
  // Determine the default branch
  if (!settings.ref && !settings.commit) {
    core.startGroup('Determining the default branch')
    settings.ref = await githubApiHelper.getDefaultBranch(
      settings.authToken,
      settings.repositoryOwner,
      settings.repositoryName
    )
    core.endGroup()
  }
  :
  :
  :
  const authHelper = gitAuthHelper.createAuthHelper(git, settings)
  try {
    // Configure auth
    core.startGroup('Setting up auth')
    await authHelper.configureAuth()
    core.endGroup()
```

phiresky:
Same here. This is breaking all of our test runs since they fetch info from a different repo than the tested one. Pretty sure this is caused by https://github.com/actions/checkout/pull/278 @ericsciple 

Using the repo token is not a valid solution in general since it doesn't work for ssh keys that are passed via secrets that are not part of the normal TOKEN.


[...]

PR #284, opened 2020-06-18
fix default branch for .wiki and when using ssh

ericsciple:
Fixes https://github.com/actions/checkout/issues/283


husss200tdd:
الف الف شكر

Hnaay-Thyy (commented):
FB_IMG_1616510231876

PR #302, opened 2020-07-13
Add deploy key scenario to README

odlp:
Hello,

This PR documents how to use a deploy key to clone an additional private repo, as an alternative to using a personal access token. 

The upside is deploy keys can be scoped to single repository with read-only access, whereas personal access tokens with the `repo` scope have read/write access to all public/private repositories of the user.

Many thanks,
Oli


odlp:
Happy 6th Birthday to this PR! 🎂🎂🎂🎂🎂🎂

PR #323, opened 2020-08-09
Add fetchJobs option to parallelize submodule updates

trylle:
This implements an option to configure the git submodule update [--jobs](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---jobsltngt) parameter.

From my testing, on a checkout with many submodules (e.g. [boost](https://www.boost.org/)) that normally takes 4.5 minutes, using this PR and declaring `fetch-jobs: 4` in the checkout step shaves off at least a minute on the checkout.


action.yml
```diff
@@ -56,6 +56,9 @@ inputs:
   fetch-depth:
     description: 'Number of commits to fetch. 0 indicates all history for all branches and tags.'
     default: 1
+  fetch-jobs:
```
staabm, line 59: Any reason to keep the default to `0`?

Any downsides of running this im parallel per default?

trylle: I've tested a bit with --jobs=0, and this probably isn't a good option for GitHub (but for the opposite reason I was worried about), since the runners are specced with only 2 cores. Since git is bottlenecked waiting for IO, you would want more jobs than that.

I think the ideal number is going to depend on a lot of factors, so it's probably better if this remains a manually enabled optimization option.

Felixoid:
Is there any reason preventing it from being merged? We have a big number of submodules in our project, and having parallel fetching would improve the time consuming *signiificantly*

Felixoid:
Would it help if I'd redo the PR from scratch?

Felixoid:
Dear @quisse and @staabm, can you kindly take a look?

quisse:
@Felixoid I'm not using github actions anymore, sorry!

staabm:
I am not someone who can push this PR either.

Felixoid:
Dear @ericsciple and @thboop, I see you've merged PRs recently. Can you review it, please?

Felixoid:
Dear @TingluoHuang and @ethomson, can you, maybe, review this extremely useful PR?

Skylion007:
@johnsudol Would you mind taking a look at this PR?

Skylion007 (approved):

src/input-helper.ts
```diff
@@ -89,6 +89,13 @@ export async function getInputs(): Promise<IGitSourceSettings> {
   }
   core.debug(`fetch depth = ${result.fetchDepth}`)
 
+  // Fetch jobs
```
johnsudol, line 92: I think this logic should be moved below `line 114`

then we can do something like: 
~~~typescript
if(result.submodules){
    result.fetchJobs = Math.floor(Number(core.getInput('fetch-jobs') || '-1'))
    if (isNaN(result.fetchJobs) || result.fetchJobs < -1) {
        etc...
}
~~~

johnsudol (commented):
Hi @trylle, thanks for the PR! I have one small comment

Based on the docs you linked, the `fetch-jobs` option will only work as an optimization when the `checkout` action is used on a repo that includes submodules, correct?

If that is the case, we only need to get the input for `fetch-jobs` if `result.submodules` is enabled.

Aside from the above, Looks good to me!

src/input-helper.ts
trylle: Applied in d46e61f

trylle:
> Based on the docs you linked, the `fetch-jobs` option will only work as an optimization when the `checkout` action is used on a repo that includes submodules, right?

As implemented, this is correct - it has no effect otherwise. It is possible to specify the `--jobs` argument to `git-fetch`, as well, but, as I understand it, its usefulness there is limited unless fetching from (potentially) multiple remotes with submodules in one go (as opposed to the fetch-then-submodule-update behavior of the `checkout` action), and would require a larger change.

Skylion007:
It seems everyone agrees on merging this PR, could we just have someone with write access take a look into approving / merging this? Ping @johnsudol . @cory-miller I also see you have merged PRs into this repo recently, would you mind taking a look?

EduMenges:
Any updates?

saagarjha:
Considering #2194 I'm guessing the chances of this getting merged are pretty slim :(

PR #326, opened 2020-08-11
Add `Licensed` To Help Verify Prod Licenses

thboop:
GitHub has a tool called [licensed](https://github.com/github/licensed) which helps us to verify that the node modules we are using are appropriately licensed for what we are using them for. It also helps to verify that the license a node module claims to be under matches the license it.

If you were previously checking in a license in the dist file, this can replace that flow.

This PR adds:
- A workflow to check licenses on pull requests and pushes to the `main` branch
- A `licensed.yml` file used to configure `licensed`
- A number of files into the `.licenses` directory which contain our dependencies and their appropriate licenses

## How does this impact me?
- You may need to [locally install](https://github.com/github/licensed#installation) licensed and run `licensed cache` to update the dependency cache if you install a new production dependency. 
  - If `licensed cache` is unable to determine the dependency, you may need to modify the cache file yourself to put the correct license. 
- You should still verify the dependency, licensed in a tool to help, but is not a substitute for [human review of dependencies](https://github.com/github/licensed#disclaimer)
- Currently, this PR only targets production dependencies, dev dependencies are not included.


ericsciple (approved):

AnandChowdhary:
@thboop, quick question: Why did you not use [jonabc/setup-licensed](https://github.com/jonabc/setup-licensed) and opt for installing Licensed manually in the workflow?

.github/workflows/licensed.yml
```diff
@@ -0,0 +1,20 @@
+name: Licensed
+
```
mary-doc, line 2: MIT

mary-doc (commented):
MIT

thorinb1983 (commented):
https://github.com/actions/checkout/pull/326/commits/3596fc0e42fd85f95cdd971e0b6d080b56478214#diff-f5b1c2f839b522a7f965389c9cd195698e98d8925964799fb3ce09804244be9f

PR #392, opened 2020-11-20
Added separate parameter for submodules fetch depth

vvish:
Using the same value of the fetch depth in the main repository and submodules can lead to a problem that cloned submodule will not contain commit referenced in the main repository. If the depth is not enough and there are commits in the submodule repository on top of the referenced one, the submodule update can fail with the following output:
```
/usr/bin/git submodule sync
/usr/bin/git -c protocol.version=2 submodule update --init --force --depth=1
Submodule '<some-submodule>' (https://<submodule>.git) registered for path '<submodule-path>'
Cloning into '/home/runner/work/<sobmodule-path>'...
Error: error: Server does not allow request for unadvertised object <some-sha>
Fetched in submodule path '<sobmodule-path>', but it did not contain <some-sha>. Direct fetching of that commit failed.
Error: The process '/usr/bin/git' failed with exit code 1
```
At the same time increasing general fetch depth will increase fetch timing for the main repository.
The proposal is to separate the parameters by introduction of the 'submodules-fetch-depth' parameter that will default to 
fetch-depth value if not present.

### DIFF
--- a/__test__/git-auth-helper.test.ts
+++ b/__test__/git-auth-helper.test.ts
@@ -763,6 +763,7 @@ async function setup(testName: string): Promise<void> {
     lfs: false,
     submodules: false,
     nestedSubmodules: false,
+    submodulesFetchDepth: 1,
     persistCredentials: true,
     ref: 'refs/heads/main',
     repositoryName: 'my-repo',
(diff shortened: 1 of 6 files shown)


ixti:
Is there any chance this will become a reality?

Issue #354, opened 2020-09-24
Getting actions workflow failed at checkout stage with error


ishuallawadhi:
: fatal: No url found for submodule path 'xyz' in .gitmodules   Error: The process '/usr/bin/git' failed with exit code 128

tchelovilar:
Did you managed to solve? I tried execute:
git rm --cached <module_path>

But I got the error again.

huihuiy02:
I have the same issue.  Any ideas/

tchelovilar:
@huihuiy02 Yes, I managed to fix removing the submodule:
`git rm --cached <path_to_submodule>`

And push the changes

But the error still happening in the self runner then I needed to remove the workflow folder from `_work` dir in the github runner home dir.

huihuiy02:
Thanks, that works for me.

cosimo:
> But the error still happening in the self runner then I needed to remove the workflow folder from `_work` dir in the github runner home dir.

This seems to be related to the error for us. We are using self-hosted runners.
Wiping the entire `_work` directory "cures" the issue, though I'm not sure what the problem is. I will be looking into this.

cosimo:
Would this be a suitable patch for `src/git-auth-helper.js`? It seems it would possibly solve the problem.

```
diff --git a/src/git-auth-helper.ts b/src/git-auth-helper.ts
index 291d83b..6783f47 100644
--- a/src/git-auth-helper.ts
+++ b/src/git-auth-helper.ts
@@ -329,22 +329,24 @@ class GitAuthHelper {
 
   private async removeGitConfig(
     configKey: string,
     submoduleOnly: boolean = false
   ): Promise<void> {
     if (!submoduleOnly) {
       if (
         (await this.git.configExists(configKey)) &&
         !(await this.git.tryConfigUnset(configKey))
       ) {
         // Load the config contents
         core.warning(`Failed to remove '${configKey}' from the git config`)
       }
     }
 
-    const pattern = regexpHelper.escape(configKey)
-    await this.git.submoduleForeach(
-      `git config --local --name-only --get-regexp '${pattern}' && git config --local --unset-all '${configKey}' || :`,
-      true
-    )
+    if (settings.submodules) {
+      const pattern = regexpHelper.escape(configKey)
+      await this.git.submoduleForeach(
+        `git config --local --name-only --get-regexp '${pattern}' && git config --local --unset-all '${configKey}' || :`,
+        true
+      )
+    }
   }
 }
```

[...]

Issue #385, opened 2020-11-11
actions/checkout@v2 breaks on a (self-hosted runner, repository) pair with submodules; no way to revert


yeputons:
How to reproduce:

1. Spin off a [self-hosted GitHub Actions runner](https://developer.github.com/v3/actions/self-hosted-runners/). This is important because for reasons I do not understand or accept (to put it mildly) self-hosted runners are fully stateful and do not clean up in any ways between runs for a specific repository.
2. Ensure that an arbitrary workflow containing `actions/checkout@v2` succeeds. Let's say that you have it in a repository `main-repo`.
3. Accidentally clone another random repository `other-repo` inside `main-repo`, and then `git add . && git commit -m"" && git push`, effectively adding `other-repo` as a submodule to `main-repo`.
4. Ensure that the workflow runs on push the first time only.
5. Re-run. Ensure that it fails like this (here `labs-yeputons` is `main-repo`, and `labs` is the `other-repo` that was accidentally add as a submodule):

```
Run actions/checkout@v2
  with:
    repository: hse-spb-2020-cpp/labs-yeputons
    token: ***
    ssh-strict: true
    persist-credentials: true
    clean: true
    fetch-depth: 1
    lfs: false
    submodules: false
Syncing repository: hse-spb-2020-cpp/labs-yeputons
Getting Git version info
  Working directory is '/home/platypus/git/actions-runner/_work/labs-yeputons/labs-yeputons'
  /usr/bin/git version
  git version 2.25.1
/usr/bin/git config --local --get remote.origin.url
https://github.com/hse-spb-2020-cpp/labs-yeputons
Removing previously created refs, to avoid conflicts
  /usr/bin/git rev-parse --symbolic-full-name --verify --quiet HEAD
  refs/heads/lab03-list-heap
  /usr/bin/git checkout --detach
  HEAD is now at d5cced7 break everything by adding submodule
  /usr/bin/git rev-parse --symbolic-full-name --branches
  refs/heads/lab03-list-heap
  /usr/bin/git branch --delete --force lab03-list-heap
  Deleted branch lab03-list-heap (was d5cced7).
  /usr/bin/git rev-parse --symbolic-full-name --remotes=origin
  refs/remotes/origin/lab03-list-heap
Cleaning the repository
  /usr/bin/git clean -ffdx
  /usr/bin/git reset --hard HEAD
  HEAD is now at d5cced7 break everything by adding submodule
Disabling automatic garbage collection
[...]

PR #424, opened 2021-01-13
GH#354 Avoid auth configuration for submodules when settings.submodules is false 

cosimo:
This patch appears to solve the problems detailed in #354 and possibly #385.
The question is: is this the correct way to solve the problem? Not sure.

The conditions to reproduce the problem are the following:
* use self-hosted github runners (in our case within docker containers, I have no idea if that's a general way to do things)
* have a private (org-wide) shared actions repository in addition to marketplace actions like `actions/checkout`, etc...
* use the workaround detailed in https://github.community/t/github-action-action-in-private-repository/16063/43 to checkout a private org-wide actions, thus using two `actions/checkout@v2` steps, one for the main project and one for the private shared actions repository.

Something happens in the "Setting up auth" stage: the additional directory used for the private shared actions repository checkout within the existing project checkout (whatever that directory is called), is interpreted as a submodule (?) when running the auth helper `removeGitConfig()` and this results in the following error messages:

```
/usr/bin/git submodule foreach --recursive git config --local --name-only --get-regexp 'core\.sshCommand' && git config --local --unset-all 'core.sshCommand' || :
  Error: fatal: No url found for submodule path '.github/my-org-github-actions' in .gitmodules
```

Note that no repository involved has submodules here.


cosimo:
We've found a good workaround, which is to simply add the directory name for the private actions checkout to `.gitignore`.

jsoref:
Please add `fixes #354` to the description

jsoref:
I'm running with this merged into https://github.com/check-spelling/actions-checkout/releases/tag/v3 as part of https://github.com/check-spelling/check-spelling/commit/407076ce6b63ad13212ef233a146c5a1bd479b94 in order to resolve the problem that this PR fixes in repositories such as https://github.com/check-spelling/winget-cli/actions/runs/3570503478/attempts/1 -- with this applied, things are happy: https://github.com/check-spelling/winget-cli/actions/runs/3571395765/jobs/6003227245#step:2:1

I expect to leave this tag in place (and possibly pull in other things as needed) until it's merged into actions/checkout@v3 ...

@cory-miller: what would it take to move this along?

cory-miller:
Can you incorporate the fix from https://github.com/actions/checkout/pull/964 to address the merge conflict? Other than that LGTM.

jsoref:
@cory-miller, I don't have write access to this PR, but I'm hosting a rebased version of the changes here: https://github.com/actions/checkout/commit/c12e7286cb39113c32dd06ac576ed748853ee61e, you should be able to force push the branch to it.

I've moved https://github.com/check-spelling/actions-checkout/releases/tag/v3 to that commit.

PR #465, opened 2021-03-24
Set initial-branch (#427)

### LINKED ISSUE #427: Suppress hint: Using 'master' as the name for the initial branch. This default branch name
Downstream consumers of actions shouldn't see noise about that git itself is changing its default branch. When `actions/checkout` is used, it will be given a specific branch (or similar) to check things out...

```
Run actions/checkout@v2.0.0
  with:
    fetch-depth: 2
    repository: org/repo
    token: ***
    persist-credentials: true
    clean: true
    lfs: false
Syncing repository: org/repo
Working directory is '/home/runner/work/org/repo'
/usr/bin/git version
git version 2.30.0
Deleting the contents of '/home/runner/work/org/repo'
/usr/bin/git init /home/runner/work/org/repo
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /home/runner/work/org/repo/.git/
```

jsoref:
Fixes #427

### DIFF
--- a/src/git-command-manager.ts
+++ b/src/git-command-manager.ts
@@ -229,7 +229,11 @@ class GitCommandManager {
   }
 
   async init(): Promise<void> {
-    await this.execGit(['init', this.workingDirectory])
+    await this.execGit([
+      'init',
+      '--initial-branch=silence_warning_about_default_branch',
+      this.workingDirectory
+    ])
   }
 
   async isDetached(): Promise<boolean> {
--- a/dist/index.js
+++ b/dist/index.js
@@ -5859,7 +5859,11 @@ class GitCommandManager {
     }
     init() {
         return __awaiter(this, void 0, void 0, function* () {
-            yield this.execGit(['init', this.workingDirectory]);
+            yield this.execGit([
+                'init',
+                '--initial-branch=silence_warning_about_default_branch',
+                this.workingDirectory
+            ]);
         });
     }
     isDetached() {


jsoref:
@ericsciple: is there something I need to do to get this considered?

jsoref:
@thboop @dscho: is there something I need to do to move this forward?

totongcute (approved):

src/git-command-manager.ts
```diff
@@ -229,7 +229,11 @@ class GitCommandManager {
   }
 
   async init(): Promise<void> {
-    await this.execGit(['init', this.workingDirectory])
+    await this.execGit([
+      'init',
+      '--initial-branch=silence_warning_about_default_branch',
```
dscho, line 234: Let's see.

First of all, I have a slight preference to `-b`, and I have a strong preference for `main` instead of that very long branch name.

However, even more important than a short & sweet name is an explanation why this won't break anything. That's really important, and I do not see anything in that regard.

Which means that the most tedious part of this PR is left to the reviewer, which is not really a good time balance.

So let's see where that analysis leads us. According to the the web interface, there is a single caller of that `init()` function, and it is [here](https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/git-source-provider.ts#L85), in context: https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/git-source-provider.ts#L80-L88

That context is _still_ not large enough to see who the branch name might be used (if at all).

A little further down, the default branch is determined: https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/git-source-provider.ts#L106-L117

Even if we moved this before the `init()` call, we could not use it as branch name because `settings.commit` might be set, in which case we're not even on a named branch (and you cannot initialize a Git repository without an initial branch name), and also `settings.ref` might refer to a tag, in which case we _also_ do not have a branch name to work with.

But let's read on.

Here is where the check out _actually_ happens: https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/git-source-provider.ts#L166-L169

That `checkoutInfo.ref` comes from https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/git-source-provider.ts#L148-L155

So let's follow the code to see what that function provides. Here is the code that sets that `ref` attribute: https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L29-L60

In particular, the `ref` attribute is set at the following locations:
- https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L31
- https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L36
- https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L42
- https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L46
- https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L51
- https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/ref-helper.ts#L54

Most importantly, the `ref` attribute is always set (unless the function throws an exception, in which case the `checkout()` call is not even reached).

Therefore, the initial branch name does not matter (as your patch suggests).

dscho (approved):
I would prefer a shorter name, and I would have _much_ preferred a proper analysis why this change is safe (instead, I was forced to perform that analysis myself, from scratch, and I imagine that this is the problem that prevents the team from merging this PR: it just asks for too much effort on the reviewers' side).

Having said that, I would not mind this PR being merged as-is, I would merely _prefer_ above-mentioned issues to be addressed.

src/git-command-manager.ts
jsoref: I'm happy to use a shorter name. I didn't want to bias a name and confuse people into thinking it mattered, as it doesn't, which is why I picked the really long one. I'm not remotely attached to it.

Would you want more detail in a comment or in a commit message? The file is fairly terse w.r.t. comments, although it would be quite reasonable to include one here (at the risk of it becoming stale if code moves).

I used long form for flags because that seemed to be house style (and to an extent the branch name because everything seems to use long descriptive names):
https://github.com/actions/checkout/blob/b342ec4a9ce8e496606d799796b72dc74766ada5/src/git-command-manager.ts#L161-L163
https://github.com/actions/checkout/blob/b342ec4a9ce8e496606d799796b72dc74766ada5/src/git-command-manager.ts#L174
https://github.com/actions/checkout/blob/b342ec4a9ce8e496606d799796b72dc74766ada5/src/git-command-manager.ts#L177-L185
...

dscho: > Would you want more detail in a comment or in a commit message?

I was thinking about the commit message, but potentially the "house style" (I like that expression) would prefer it in the PR description. Please note that I am _not_ on the team that develops this GitHub Action, therefore I have no power to merge or reject it. Read: my opinion does not matter _all_ that much here ;-)

> I used long form for flags because that seemed to be house style

Fair enough. When in Rome...

jsoref:
@hross what do I need to do to move this forward?

dscho:
@ericsciple @TingluoHuang ping?

jsoref:
@vanZeben is there something that can be done to move this along?

kbrock:
Looks like this stalled.

Are there other suggestions on silencing this warning? It shows up quite a few times.

MestreLion:
My opinions on this:
- The long form `--initial-branch` was a great choice. It's self-descriptive, and would have little to no point saving a few bytes with `-b`
- The chosen branch name is only temporary and not referenced anywhere. It couldn't be otherwise, as the current `master` used by default is not hardcoded anywhere in this tool, and even when it doesn't match the cloned repo's this tool works on them just fine.
- So, how about `dummy`? It's short, it clearly states it is temporary and irrelevant, and it avoids taking sides about `master`/`main`.

And remember that just like @dscho I'm not a maintainer (and not even a contributor), just a developer very interested in this PR going forward and eventually being accepted. It's been more than an year guys!

(PS: @jsoref , you might mark your conversation with @dscho as _**resolved**_, it would be one less red status on the pending items summary, this _might_ incentivize a reviewer . And maybe ping an actual (human) maintainer/reviewer, as that `CODEOWNERS` file you used to ping the team seems to have errors (`@actions/actions-runtime` is not an actual team), so no one was actually pinged.)

jsoref:
@mouismail?

I'm happy to pick any name that a codeowner suggests.

Personally I use `placeholder` for my default branches (it doesn't have any negative connotations), but I don't want people to think that I'm actively favoring that here or that my use of `placeholder` would relate to the use here...

Anyway, I'm more than happy to update this (or any of my other PRs) although I'm about to be relatively unavailable for a bit.

MestreLion:
> Anyway, I'm more than happy to update this (or any of my other PRs) although I'm about to be relatively unavailable for a bit.

I suggest you preemptively change the branch name now to something shorter, such as your `placeholder` (a good choice!), or `dummy`, `main` or even `master` (current implicit default), just to avoid a situation where a reviewer asks for a change and this PR blocks again because you're unavailable. I believe the shorter names above will make it more likely for this PR to be accepted and merged without changes even if you're AFK.

willrowe:
Git now includes this at the end of the message:
>  hint: Disable this message with "git config set advice.defaultBranchName false"

It seems that setting that config value would accomplish suppressing the message in the most direct way.

jsoref:
No.

https://github.com/git/git/blame/v2.50.0/t/t0001-init.sh#L833-L837 requires v2.50.0 and many runners don't have that.

Debian 12.8 has `git version 2.39.5`

jsoref:
Again, it's really pointless for me to do *anything* to a PR to *any* github owned repo until a GitHub staffer shows up and says "hey, I'm willing to talk to you about this."

This repository appears to be mostly abandoned.

@TingluoHuang is welcome to prove me wrong.

willrowe:
@jsoref the config value can be set regardless of whether the installed version supports it or not. It would then help with newer versions going forward. I suggested this as an alternative that would not require any bikeshedding over the default branch name.

jsoref:
That's totally useless. And there's no point in bikeshedding. This PR works for any version of git that has the warning. It's only useful to bikeshed if there's a living maintainer. And there aren't.

```sh
root@0db61f84061c:/tmp/tmp.1sO90FTg4c# git version
git version 2.39.5
root@0db61f84061c:/tmp/tmp.1sO90FTg4c# git -c advice.defaultBranchName=false init
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint: 	git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint: 	git branch -m <name>
Initialized empty Git repository in /tmp/tmp.1sO90FTg4c/.git/
```

dscho:
@jsoref using an absurdly long branch name can run into "long path" problems on Windows, especially with recursive clones.

If I were a maintainer and saw the resistance against implementing obvious improvements, even if I had the time to engage I probably wouldn't unless someone forced me to.

MestreLion:
> If I were a maintainer and saw the resistance against implementing obvious improvements, even if I had the time to engage I probably wouldn't unless someone forced me to.

What resistance? Most (if not all) comments _approved_ the PR. Any maintainer can just ignore the bikeshedding about the branch name, as many (including myself and the author) said any name is fine and should not be a blocker.

dscho:
> and should not be a blocker.

Well, I have told you that this long name _is_ a blocker, as it can cause issues with long paths on Windows when cloning recursively.

And as I have said before, if I were a maintainer, I wouldn't appreciate it if any contributor simply tried to ignore away a valid concern like this.

love1120520 (approved):

PR #476, opened 2021-04-11
chore: create `dependabot.yml`

MichaelDeBoey:
This will create automatic PRs to update both dependencies + actions used in workflows


cory-miller:
There were a few versions of this config in various PRs that I combined into https://github.com/actions/checkout/pull/1688

PR #480, opened 2021-04-18
Add required required field to action.yml

jsoref:
https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_idrequired

> inputs.<input_id>.required
Required A boolean to indicate whether the action requires the input parameter. Set to true when the parameter is required.


jsoref:
@hross is there something I need to do to move this along?

PR #482, opened 2021-04-23
Improve docs for checking out a private repo

nholden:
I tried using `actions/checkout` to checkout a private repo using the README, and I bumped into two issues:

1. I thought that `secrets.GitHub_PAT` was something that I got for free. It turns out that the user must create this secret themselves.
2. User-created secrets [cannot start with `GitHub_`](https://docs.github.com/en/actions/reference/encrypted-secrets).

This change updates the README to clarify both of those points.


nholden:
Fixed by https://github.com/actions/checkout/pull/623

PR #561, opened 2021-08-03
Don't erase the repository directory when clean is false and the repo isn't cloned

dra27:
At present, if the repository directory doesn't contain a `.git` directory then the action always erases the entire repository directory, regardless of the "clean" setting. This PR amends the behaviour so that for `clean: false` the directory is _not_ erased if there's no `.git` directory (if there is a `.git` directory, then the checks proceed as before). It's obviously the user's responsibility (having set `clean: false`) to ensure that the files left in the directory don't interface with ones which would be checked out.

This was hit in ocaml/setup-ocaml@v2 (see ocaml/setup-ocaml#190). setup-ocaml v2 does some work in `_opam` which was then being unnecessarily erased if actions/checkout was used in a later step.


smorimoto:
@thboop @ericsciple Could you check this?

smorimoto:
CC: @dhadka

src/git-directory-helper.ts
```diff
@@ -24,10 +24,10 @@ export async function prepareExistingDirectory(
     remove = true
   }
   // Fetch URL does not match
-  else if (
-    !fsHelper.directoryExistsSync(path.join(repositoryPath, '.git')) ||
-    repositoryUrl !== (await git.tryGetFetchUrl())
-  ) {
+  else if (!fsHelper.directoryExistsSync(path.join(repositoryPath, '.git'))) {
```
tomcashman, line 27: I believe this should remove the `!`

PR #584, opened 2021-09-15
Set default user.name and user.email

### LINKED ISSUE #13: Set git user and email

jablko:
This PR sets up a default `user.name` and `user.email` for scripting `git commit`. Users can continue to configure their own values as now, this PR just affords the option not to.

Shouldn't affect existing users: If users `git config --global user.{name,email}` before `actions/checkout`, this will respect existing values. If they `git config --global` after `actions/checkout` they'll overwrite these defaults. If they `git config --local` (before or after), those values take precedence.

Fixes #13, although the issue proposed defaulting to the author/user from the event that triggered the action.

Also discussed in #158, although the PR doesn't propose defaults.

Configuring `user.{name,email}`/scripting `git commit` is documented in the [README.md](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token). The relevant text of [ADR 0153](https://github.com/actions/checkout/blob/main/adrs/0153-checkout-v2.md#persist-credentials) reads:
> Note:
> - Users scripting `git commit` may need to set the username and email. The service does not provide any reasonable default value. Users can add `git config user.name <NAME>` and `git config user.email <EMAIL>`. We will document this guidance.

### Default user.name

I used [`$GITHUB_WORKFLOW`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#:~:text=the%20name%20of%20the%20workflow.%20if%20the%20workflow%20file%20doesn't%20specify%20a%20name%2C%20the%20value%20of%20this%20property%20is%20the%20full%20path%20of%20the%20workflow%20file%20in%20the%20repository.) for the default `user.name` because it's the most helpful value to have in there? Other proposals include the current [README.md](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token) value ("github-actions") and the [author/user from the event that triggered the action](https://github.com/actions/checkout/pull/158#discussion_r379766291).

Here's how `$GITHUB_WORKFLOW` looks in GitHub's UI -- however I'm equally happy if a different default is implemented: Screenshot from 2021-09-10 10-55-25

### Default user.email

I went with the current [README.md](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token) value ("github-actions@github.com"), however @JojOatXGME [lays out the options](https://github.com/actions/checkout/issues/13#issuecomment-724415212).

### DIFF
--- a/src/git-source-provider.ts
+++ b/src/git-source-provider.ts
@@ -1,4 +1,5 @@
 import * as core from '@actions/core'
+import * as github from '@actions/github'
 import * as fsHelper from './fs-helper'
 import * as gitAuthHelper from './git-auth-helper'
 import * as gitCommandManager from './git-command-manager'
@@ -216,6 +217,14 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
       settings.ref,
       settings.commit
     )
+
+    // Set default author
+    if (!await git.configExists('user.name', true) {
+      await git.config('user.name', github.context.workflow, true)
+    }
+    if (!await git.configExists('user.email', true) {
+      await git.config('user.email', 'github-actions@github.com', true)
+    }
   } finally {
     // Remove auth
     if (!settings.persistCredentials) {


src/git-source-provider.ts
```diff
@@ -216,6 +217,14 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
       settings.ref,
       settings.commit
     )
+
+    // Set default author
+    if (!await git.configExists('user.name', true) {
+      await git.config('user.name', github.context.workflow, true)
+    }
+    if (!await git.configExists('user.email', true) {
```
usmonster, line 225:
```suggestion
    if (!await git.configExists('user.name', true)) {
      await git.config('user.name', github.context.workflow, true)
    }
    if (!await git.configExists('user.email', true)) {
```

jablko: Oops, thank you!

src/git-source-provider.ts
```diff
@@ -216,6 +217,14 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
       settings.ref,
       settings.commit
     )
+
+    // Set default author
+    if (!await git.configExists('user.name', true)) {
+      await git.config('user.name', github.context.workflow, true)
+    }
+    if (!await git.configExists('user.email', true)) {
+      await git.config('user.email', 'github-actions@github.com', true)
```
NatoBoram, line 226: There's already a bot account for GitHub Actions. It seems like it's a decent default.

https://github.com/actions/checkout/blob/692973e3d937129bcbf40652eb9f2f61becf3332/README.md?plain=1#L283-L284

```suggestion
      await git.config('user.name', 'github-actions[bot]', true)
    }
    if (!await git.configExists('user.email', true)) {
      await git.config('user.email', '41898282+github-actions[bot]@users.noreply.github.com', true)
```

qoomon:
Maybe it make sense to add user as an action input and choose who should be set as user e.g. bot, actor, committer or even manually

I've created a handy action for that use case. Feel free to have look at https://github.com/qoomon/actions--setup-git

farshadjanu1 (approved):

PR #620, opened 2021-10-20
codeql should analyze lib not dist

ericsciple:


thboop (approved):
lgtm. We should see if we can request the codeql action take an allowlist/ignorelist if possible

adityasharad:
👋  Greetings from the CodeQL team! Just spotted this in passing and wanted to offer a suggestion.
CodeQL's JavaScript analysis can do TypeScript parsing, so I think you could omit the build step entirely in your Actions workflow, and it will analyse the `.ts` files. The only thing you'll miss out is code that is generated only during the build (if any).

If you still need to filter out directories from being scanned, see [these docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan).

PR #623, opened 2021-10-24
Change secret name for PAT to not start with GITHUB_

juggernaut:
Github doesn't allow secret names that start with `GITHUB_` (case insensitive). Update README to choose a different prefix (GH).


ericsciple:
thx!

PR #651, opened 2021-12-08
Check for symlinked gitconfig

msdrigg:
We don't want to edit the user's actual gitconfig, we only want to copy their setup.

If the file is symlinked, we resolve the link and copy the source file instead.


msdrigg:
Any thoughts on this?

Kevin-Li-Webjet:
looks better to me

msdrigg:
Can we get this merged then

chrisoro:
I encountered the same issue. Would it be possible to merge this pull request?

chrisoro:
@jww3 any chance we can get this in?

daltonbr:
I also had the same issue.

vasconcelosvcd:
this is till happening and it's a big issue on golang projects which relies on the git config so private modules can be properly downloaded

PR #656, opened 2021-12-14
chore: setup Dependabot for Actions and NPM

nschonni:
Not sure if some might require manual intervention with the license check, but at least it will start to get deps update


cory-miller:
There were a few versions of this config in various PRs that I combined into https://github.com/actions/checkout/pull/1688

PR #679, opened 2022-01-27
Updating readme

romulofreires1:
Adding "# Default: ${{ github.ref }}" in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous
     # The branch, tag or SHA to checkout. When checking out the repository that
     # triggered a workflow, this defaults to the reference or SHA for that event.
     # Otherwise, uses the default branch.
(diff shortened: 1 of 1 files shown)


Wrathchyld:
steps:
- uses: actions/checkout@v2
- uses: actions/setup-node@v2
  with:
    node-version: '14'
    cache: 'npm'
- run: npm install
- run: npm test

cory-miller:
The README should by synced with `action.yml`.

https://github.com/actions/checkout/blob/a5ac7e51b41094c92402da3b24376905380afc29/action.yml#L7-L11

Note that the default annotations in `action.yml` is not just reference documentation. So this PR actually changes the behavior seen in https://github.com/actions/checkout/blob/a5ac7e51b41094c92402da3b24376905380afc29/src/input-helper.ts#L59-L72 by guaranteeing that the input is set (and never entering the if statement). I don't think we should to change this behavior.

Issue #172, opened 2020-03-02
Options like sparse mode


jjangga0214:
Hi. 

GitHub is internally evaluating [partial-clone](https://git-scm.com/docs/partial-clone) (Actually GitLab already provides the feature!).

This can be especially useful for large monorepo, with *sparse-checkout*(see `git sparse-checkout --help`) feature.

Thus, hereby I request support to those features.

Thanks.

**workaround**

Here is a workaround suggestion for folks who want to customize the whole process, though I haven't tested.

```bash
$ REPO="https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"

$ git clone $REPO --your-options
```

annulen:
FWIW, Travis CI also supports sparse checkout, described at https://docs.travis-ci.com/user/customizing-the-build#git-sparse-checkout

It would be even better to have full-fledged partial-clone though.

ericsciple:
`actions/checkout@v2` is optimized to fetch only a single commit by default, so you should get some of the same benefits (no unnecessary history)

However doesn't solve the files-outside-of-the-user’s-work-area-in-the-tree problem.

annulen:
It might be a good idea to add `no-checkout` option for action, so that people could customize checkout command directly without reimplementing initial setup

annulen:
For now I'm using this
```
    - name: Sparse checkout
      shell: bash
      run: |
        REPO="https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
        BRANCH="${GITHUB_REF/#refs\/heads\//}"

        # Following code is based on logs of actions/checkout@v, with sparseCheckout stuff inserted in the middle
        echo "Syncing repository: $GITHUB_REPOSITORY"
        echo "Working directory is '$(pwd)' GITHUB_WORKSPACE=$GITHUB_WORKSPACE BRANCH=$BRANCH"
        git version
        git init $GITHUB_WORKSPACE
        git remote add origin https://github.com/$GITHUB_REPOSITORY
        git config --local gc.auto 0
        # Now interesting part
        git config core.sparseCheckout true
        # Add here contents of sparse-checkout line by line
        echo "..." >> .git/info/sparse-checkout
        # echo ... 
        git -c protocol.version=2 fetch --no-tags --prune --progress --depth=10 origin +${GITHUB_SHA}:refs/remotes/origin/${BRANCH}
[...]

PR #680, opened 2022-02-01
Add support for sparse checkouts

jonnelaaglandwinder:
I've taken a crack at implementing a sparse checkout option suggested in #172.
It may prove useful for large monorepos, especially if a --filter option for fetch were to be implemented as well.

I've put a temporary escape hatch in the test script, since it would fail without the test data branch being present, but it should pass with those lines removed when the test data exists. ([test-data/v2/sparse-checkout](https://github.com/Nocturne92/checkout/tree/test-data/v2%2Fsparse-checkout))


__test__/verify-sparse-checkout.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+## REMOVE THIS
+exit 0
+## REMOVE THIS
```
rcowsill, line 5: Should this be removed now?

arcanis:
Bump @thboop? This would be extremely useful for repositories that check-in (somewhat) large artifacts, like their zipped dependencies, which is a somewhat common pattern with Yarn.

Wrathchyld:
action.yml

tyson-benson:
This would be a great addition 👍 
It would be a valuable capability for those that want to adopt mono-repo design. I hope this PR gets the attention it deserves.

arcanis:
Ping @TingluoHuang? Are there any blockers?

__test__/verify-sparse-checkout.sh
```diff
@@ -0,0 +1,25 @@
...
+
```
austinpray-mixpanel, line 6:
```suggestion
```

blu3mania:
Shouldn't lfs.fetchinclude also be set or use "-I" (or "--include") parameter to make sure LFS only fetches from included paths?

guitarrapc:
Is any blocker exists? I'm eager to use this feature but wait for half a year.

derekperkins:
Any feedback from @actions-runtime?

hyoretsu:
This would be great for monorepos y'know.

zchenyu:
Another friendly bump on this. +1 for this feature

dfdez:
Hey everyone! I see that this PR has been frozen for a while...

In my company we need to use the `sparse-checkout` functionality, that is why I have invested time in order to retake this. 

Here you can find a [PR](https://github.com/actions/checkout/pull/1317) with an updated version to add support for the `sparse-checkout`!

I have to mention that this PR has helped me in order to make mine so thanks @jonnelaaglandwinder to start this, let's hope we can finally have this functionality implemented!

Spiritaine (approved):

dfdez:
This PR can be closed due https://github.com/actions/checkout/pull/1369

Spiritaine (approved):

Issue #290, opened 2020-06-26
Preserve tag annotations


martinthomson:
The technique this tool uses for checking out tags only preserves the relation between tag and ref, not any annotations created with `git tag -a`.

```
git fetch origin +$HASH:refs/tags/$TAG
git checkout -f refs/tag/$TAG
```

I have a CI action that depends on information from annotated tags.  Specifically, the name of the person who applied the tag: `git tag --list --points-at HEAD --format '%(taggeremail)'` (I can't use `${{ github.actor }}` for reasons).

The workaround is to run `git fetch -f origin ${{ github.ref }}:${{ github.ref }}`.  But it would be better not to need this workaround.

morrone:
Agreed! This breaks the "git describe" command too. We need tags to be in their correct original form.

martinthomson:
I realize now that this is complicated by a desire to checkout the specific code identified by the hash when the tag has been re-applied on the repository.  I would be happy to have an option that would cause the checkout to fail if the hash didn't match.

For example:
```
git fetch origin refs/tags/$TAG
[ `git rev-parse refs/tags/$TAG` != $HASH ] || fail
git checkout -f refs/tag/$TAG
```

ericsciple:
@martinthomson does the input `depth: 0` help? It will cause all history for all branches and tags to be fetched?

morrone:
> @martinthomson does the input `depth: 0` help? It will cause all history for all branches and tags to be fetched?

I didn't try 0, but I tried "fetch-depth: 2000" at one point. It fetched the _other_ tags in that history correctly, but if the thing that triggered the workflow was itself an annotated tag, that one tag will still be converted from an annotated tag into a lightweight tag.

ericsciple:
`fetch-depth: 0` triggers a different behavior. Curious if that solves the issue?

Stanzilla:
It does not solve the issue, I ran into this today.

andreineculau:
`fetch-depth: 0` does solve the issue. I don't see how it wouldn't because it fetch all branches, all tags

Stanzilla:
@andreineculau sadly does not.

Envek:
+1 on preserving tag annotations. `fetch-depth: 0` alone doesn't help, `git fetch --tags --force` does help.

```yaml
name: Create release from annotated tag
on: push
jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
[...]

PR #697, opened 2022-02-14
Don't overwrite annotated tags with commit object

bk2204:
When checking out a repository with full history, a full clone is done and then the ref is finally updated to point to the commit that caused the workflow to be run.  Normally, this is a good protection against someone pushing to the repository twice in short succession, but it causes problems with annotated tags.

Specifically, because the entry in refs/tags is set to the commit hash, if an annotated tag was used, the tag is turned merely into a lightweight one, which breaks `git describe`.  Every other tag in the repository will continue to remain a valid annotated tag except the one for which the workflow was invoked, which is not what the user expected.

Let's work around this by not performing a fetch if what we're fetching is a tag.  Technically, annotated tags can be anywhere in the hierarchy at any ref, but this should work as a suitable heuristic for now.

Note that the proper solution would be to expose the revision of the actual object and check against that instead of the commit, but it doesn't presently appear that that information is exposed.  Also, we explicitly do not case-fold since Git refs are case sensitive.

This mostly addresses #290.

### DIFF
--- a/src/git-source-provider.ts
+++ b/src/git-source-provider.ts
@@ -135,7 +135,7 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
 
       // When all history is fetched, the ref we're interested in may have moved to a different
       // commit (push or force push). If so, fetch again with a targeted refspec.
-      if (!(await refHelper.testRef(git, settings.ref, settings.commit))) {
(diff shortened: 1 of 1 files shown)


martinpitt:
I keep running into this bug. This is the standard workaround which is in all our cockpit projects, as well as my own now:

```yaml
      - name: Workaround for https://github.com/actions/checkout/pull/697
        run: git fetch --force origin $(git describe --tags):refs/tags/$(git describe --tags)
```

PR #708, opened 2022-03-01
update readme for v3

thboop:


ericsciple (approved):

ericsciple:
Do we need a section to indicate whats new in v3?

ericsciple:
Basically, what do users on v2 need to know?

kokofirstlove (commented):
``

jekyleee (approved):

eregon:
I'd be useful to mention this node update is because of, quoting https://github.com/actions/checkout/pull/689
> Node 12 has an end of life on April 30, 2022.

I'm not sure why this should be a major release, it feels like a lot of churn for every action out there, most of them likely running on GH-hosted or recent-enough runners, and for which using a new 2.x release would have been better.

Does it mean `- uses: actions/checkout@v2` which is used in so many workflows will no longer get updates/bug fixes?

This feels wrong, people using old runners should pin to a specific release or sha, people using reasonably recent runners should be able to get updates/bug fixes without needing to update all workflows with a new `checkout` major version every now and then.

README.md
```diff
@@ -14,27 +14,14 @@ When Git 2.18 or higher is not in your PATH, falls back to the REST API to downl
...
-- Improved layout
-  - The input `path` is always relative to $GITHUB_WORKSPACE
-  - Aligns better with container actions, where $GITHUB_WORKSPACE gets mapped in
-- Fallback to REST API download
-  - When Git 2.18 or higher is not in the PATH, the REST API will be used to download the files
-  - When using a job container, the container's PATH is used
-
-Refer [here](https://github.com/actions/checkout/blob/v1/README.md) for previous versions.
+- Updated to the node16 runtime by default
+  - This requires a minimum [Actions Runner](https://github.com/actions/runner/releases/tag/v2.285.0) version of v2.285.0 to run, which is by default available in GHES 3.4 or later.
```
jglick, line 18: I think it would be worth explicitly mentioning here (and, more importantly, in https://github.com/actions/checkout/releases/tag/v3.0.0) that if you are using github.com (SaaS) you can safely accept the update and ignore the rest of the text.

PR #710, opened 2022-03-02
Update changelog

bquorning:
When Dependabot updated actions/checkout to v3.0.0 in my repository, it only shows the changelog up until v2.3.1.

The text is largely copied from https://github.com/actions/checkout/releases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,28 @@
 
 - [Update to node 16](https://github.com/actions/checkout/pull/689)
 
+## v2.4.0
+
+- Convert SSH URLs like `org-<ORG_ID>@github.com:` to `https://github.com/` [PR](https://github.com/actions/checkout/pull/621)
+
+## v2.3.5
+
+- Update dependencies
+
+## v2.3.4
+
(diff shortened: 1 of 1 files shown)


jonasbn:
Alternatively the `CHANGELOG.md` file could be deprecated and removed in favor of the releases on GitHub.

Change log files have a lot of value when you are distributing your source code, but this does not really leave the GitHub eco-system - I presume.

The currently situation of releases and file not being in sync is just confusing.

Great work by @bquorning btw. but it is _hard_ work to keep things in sync.

bquorning:
> it is _hard_ work to keep things in sync.

I’m sure someone can write a GHA workflow to copy release information into the Changelog (or vice versa). Or change Dependabot to make its PR descriptions from GH releases. 😄

jonasbn:
@bquorning I could actually use that - I will sleep on it

bquorning:
Updated after #740 has been merged (long ago)

cory-miller:
I added instructions to the release playbook to update the changelog when bumping major versions, to hopefully prevent this from happening again.

PR #762, opened 2022-04-13
Enforce safe directory

### LINKED ISSUE #760: `fatal: unsafe repository (REPO is owned by someone else)` with ubuntu 20.04 container
Hello!

This started happening recently; I see it both with v2.4.0 and v.3.0.0:

```
Syncing repository: wez/wezterm
Getting Git version info
Deleting the contents of '/__w/wezterm/wezterm'
Initializing the repository
  /usr/bin/git init /__w/wezterm/wezterm
  Initialized empty Git repository in /__w/wezterm/wezterm/.git/
  /usr/bin/git remote add origin https://github.com/wez/wezterm
  Error: fatal: unsafe repository ('/__w/wezterm/wezterm' is owned by someone else)
  To add an exception for this directory, call:
  
  	git config --global --add safe.directory /__w/wezterm/wezterm
```

Is there a way to configure/advise the action how to handle this?
Is this really a problem elsewhere in the GH actions environment?

thboop:
This pr fixes #760 by setting the repositories path as a safe directory when running checkout and removing that config on cleanup.

We aren't able to set it as a local config, as this setting only works for global configs.

So, using the standard we used for submodules, lets go ahead and copy the existing global config to a new location. Then we can go ahead and modify that global config (so we don't modify like a users global config if they test something on a self hosted runner real fast). Then, we can delete that global config at the end of the checkout step.

This does carry some limitations, it doesn't persist this configuration for the duration of the job mainly. So if your job pushes to git or something after checkout, that will continue to fail. We need to figure out how to address this at an ecosystem level, outside of the checkout action.

### DIFF
--- a/__test__/git-auth-helper.test.ts
+++ b/__test__/git-auth-helper.test.ts
@@ -643,10 +643,11 @@ describe('git-auth-helper tests', () => {
     expect(gitConfigContent.indexOf('http.')).toBeLessThan(0)
   })
 
-  const removeGlobalAuth_removesOverride = 'removeGlobalAuth removes override'
-  it(removeGlobalAuth_removesOverride, async () => {
+  const removeGlobalConfig_removesOverride =
+    'removeGlobalConfig removes override'
+  it(removeGlobalConfig_removesOverride, async () => {
     // Arrange
-    await setup(removeGlobalAuth_removesOverride)
+    await setup(removeGlobalConfig_removesOverride)
     const authHelper = gitAuthHelper.createAuthHelper(git, settings)
     await authHelper.configureAuth()
     await authHelper.configureGlobalAuth()
@@ -655,7 +656,7 @@ describe('git-auth-helper tests', () => {
     await fs.promises.stat(path.join(git.env['HOME'], '.gitconfig'))
 
     // Act
-    await authHelper.removeGlobalAuth()
+    await authHelper.removeGlobalConfig()
 
     // Assert
     expect(git.env['HOME']).toBeUndefined()
--- a/dist/index.js
+++ b/dist/index.js
@@ -6572,9 +6572,13 @@ class GitAuthHelper {
             yield this.configureToken();
         });
     }
-    configureGlobalAuth() {
-        var _a;
+    configureTempGlobalConfig(repositoryPath) {
+        var _a, _b;
         return __awaiter(this, void 0, void 0, function* () {
+            // Already setup global config
+            if (((_a = this.temporaryHomePath) === null || _a === void 0 ? void 0 : _a.length) > 0) {
+                return path.join(this.temporaryHomePath, '.gitconfig');
+            }
             // Create a temp home directory
             const runnerTemp = process.env['RUNNER_TEMP'] || '';
             assert.ok(runnerTemp, 'RUNNER_TEMP is not defined');
@@ -6590,7 +6594,7 @@ class GitAuthHelper {
                 configExists = true;
             }
             catch (err) {
-                if (((_a = err) === null || _a === void 0 ? void 0 : _a.code) !== 'ENOENT') {
+                if (((_b = err) === null || _b === void 0 ? void 0 : _b.code) !== 'ENOENT') {
                     throw err;
                 }
             }
@@ -6601,10 +6605,25 @@ class GitAuthHelper {
             else {
                 yield fs.promises.writeFile(newGitConfigPath, '');
             }
+            // Override HOME
+            core.info(`Temporarily overriding HOME='${this.temporaryHomePath}' before making global git config changes`);
+            this.git.setEnvironmentVariable('HOME', this.temporaryHomePath);
+            // Setup the workspace as a safe directory, so if we pass this into a container job with a different user it doesn't fail
+            // Otherwise all git commands we run in a container fail
+            core.info(`Adding working directory to the temporary git global config as a safe directory`);
+            yield this.git
+                .config('safe.directory', repositoryPath !== null && repositoryPath !== void 0 ? repositoryPath : this.settings.repositoryPath, true, true)
+                .catch(error => {
+                core.info(`Failed to initialize safe directory with error: ${error}`);
+            });
+            return newGitConfigPath;
+        });
+    }
+    configureGlobalAuth() {
+        return __awaiter(this, void 0, void 0, function* () {
+            // 'configureTempGlobalConfig' noops if already set, just returns the path
+            const newGitConfigPath = yield this.configureTempGlobalConfig();
             try {
-                // Override HOME
-                core.info(`Temporarily overriding HOME='${this.temporaryHomePath}' before making global git config changes`);
-                this.git.setEnvironmentVariable('HOME', this.temporaryHomePath);
                 // Configure the token
                 yield this.configureToken(newGitConfigPath, true);
                 // Configure HTTPS instead of SSH
@@ -6657,11 +6676,14 @@ class GitAuthHelper {
             yield this.removeToken();
         });
     }
-    removeGlobalAuth() {
+    removeGlobalConfig() {
+        var _a;
         return __awaiter(this, void 0, void 0, function* () {
-            core.debug(`Unsetting HOME override`);
-            this.git.removeEnvironmentVariable('HOME');
-            yield io.rmRF(this.temporaryHomePath);
+            if (((_a = this.temporaryHomePath) === null || _a === void 0 ? void 0 : _a.length) > 0) {
+                core.debug(`Unsetting HOME override`);
+                this.git.removeEnvironmentVariable('HOME');
+                yield io.rmRF(this.temporaryHomePath);
+            }
         });
     }
     configureSsh() {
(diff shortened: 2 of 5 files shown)


src/git-command-manager.ts
```diff
@@ -44,6 +44,11 @@ export interface IGitCommandManager {
   tagExists(pattern: string): Promise<boolean>
   tryClean(): Promise<boolean>
   tryConfigUnset(configKey: string, globalConfig?: boolean): Promise<boolean>
+  tryConfigUnsetValue(
+    configKey: string,
+    valueRegex: string,
+    globalConfig?: boolean
+  )
```
TingluoHuang, line 51: `: Promise<boolean>`

src/git-source-provider.ts
```diff
@@ -245,6 +250,13 @@ export async function cleanup(repositoryPath: string): Promise<void> {
   // Remove auth
   const authHelper = gitAuthHelper.createAuthHelper(git)
   await authHelper.removeAuth()
+
+  // always do this last, otherwise other git commands may fail
+  await git
+    .tryConfigUnsetValue('safe.directory', `'${repositoryPath}$'`, true)
+    .catch(error => {
```
TingluoHuang, line 257: will `tryConfigUnsetValue` ever throw? or we should check the return value from `tryConfigUnsetValue()`?

src/git-source-provider.ts
```diff
@@ -245,6 +250,13 @@ export async function cleanup(repositoryPath: string): Promise<void> {
   // Remove auth
   const authHelper = gitAuthHelper.createAuthHelper(git)
   await authHelper.removeAuth()
+
+  // always do this last, otherwise other git commands may fail
+  await git
+    .tryConfigUnsetValue('safe.directory', `'${repositoryPath}$'`, true)
```
ethomson, line 256: I think that we should not use the `repositoryPath` literally within a regex. I think that we'd need to translate special characters in the path (like `.`). Do we need a start anchor `^` as well?

But - could we avoid all of that and do `git -cgrep.patternType=fixed unset ...` and treat it like a fixed string and not like a regex?

src/git-source-provider.ts
```diff
@@ -82,6 +82,11 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
     !fsHelper.directoryExistsSync(path.join(settings.repositoryPath, '.git'))
   ) {
     core.startGroup('Initializing the repository')
+    await git
```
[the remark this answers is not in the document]
thboop: Yes, fixed!

__test__/git-auth-helper.test.ts
```diff
@@ -643,10 +643,11 @@ describe('git-auth-helper tests', () => {
     expect(gitConfigContent.indexOf('http.')).toBeLessThan(0)
   })
 
-  const removeGlobalAuth_removesOverride = 'removeGlobalAuth removes override'
-  it(removeGlobalAuth_removesOverride, async () => {
+  const removeGlobalConfig_removesOverride =
+    'removeGlobalConfig removes override'
+  it(removeGlobalConfig_removesOverride, async () => {
     // Arrange
-    await setup(removeGlobalAuth_removesOverride)
```
thboop, line 649: Renamed this function, changes are a result of that

src/git-source-provider.ts
```diff
@@ -244,7 +248,12 @@ export async function cleanup(repositoryPath: string): Promise<void> {
 
   // Remove auth
   const authHelper = gitAuthHelper.createAuthHelper(git)
-  await authHelper.removeAuth()
+  try {
+    await authHelper.configureTempGlobalConfig(repositoryPath)
```
thboop, line 252: We also need to do this on cleanup, otherwise cleanup will fail

src/git-auth-helper.ts
```diff
@@ -80,7 +81,11 @@ class GitAuthHelper {
     await this.configureToken()
   }
 
-  async configureGlobalAuth(): Promise<void> {
+  async configureTempGlobalConfig(repositoryPath?: string): Promise<void> {
```
thboop, line 84: Reviewing with whitespace disabled will make this much easier to read

src/git-source-provider.ts
```diff
@@ -170,34 +179,26 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
 
-        // Persist credentials
-        if (settings.persistCredentials) {
-          core.startGroup('Persisting credentials for submodules')
-          await authHelper.configureSubmoduleAuth()
-          core.endGroup()
-        }
-      } finally {
-        // Remove temporary global config override
-        await authHelper.removeGlobalAuth()
```
thboop, line 200: We don't need to do this here, as we do it at the try finally wrapped around this entire block

cbalioglu:
@thboop thanks a lot for taking quick action! Do you have an ETA on when you plan to land this fix?

src/git-auth-helper.ts
```diff
@@ -109,14 +114,40 @@ class GitAuthHelper {
     } else {
       await fs.promises.writeFile(newGitConfigPath, '')
     }
-
```
TingluoHuang, line 112: add me back?

src/git-auth-helper.ts
```diff
@@ -109,14 +114,40 @@ class GitAuthHelper {
...
     try {
       // Override HOME
       core.info(
         `Temporarily overriding HOME='${this.temporaryHomePath}' before making global git config changes`
       )
       this.git.setEnvironmentVariable('HOME', this.temporaryHomePath)
+    } catch (err) {
```
TingluoHuang, line 123: why do we need a catch around `code.info` and `this.git.setEnvironmentVariable`?

src/git-auth-helper.ts
```diff
@@ -19,8 +19,9 @@ export interface IGitAuthHelper {
   configureAuth(): Promise<void>
   configureGlobalAuth(): Promise<void>
   configureSubmoduleAuth(): Promise<void>
+  configureTempGlobalConfig(repositoryPath?: string): Promise<void>
```
TingluoHuang, line 22: what about making `configureTempGlobalConfig` return the path to the temp config file?

TingluoHuang: so we don't need to do `const newGitConfigPath = path.join(this.temporaryHomePath, '.gitconfig')` twice?

src/git-auth-helper.ts
```diff
@@ -109,14 +114,40 @@ class GitAuthHelper {
...
+      throw err
+    }
+
+    // Setup the workspace as a safe directory, so if we pass this into a container job with a different user it doesn't fail
+    // Otherwise all git commands we run in a container fail
+    core.info(
+      `Adding working directory to the temporary git global config as a safe directory`
+    )
+    await this.git
+      .config(
```
TingluoHuang, line 134: does the `.config` handles the `--add` for you?

src/git-source-provider.ts
```diff
@@ -36,68 +36,77 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
-  if (isExisting) {
-    await gitDirectoryHelper.prepareExistingDirectory(
-      git,
-      settings.repositoryPath,
-      repositoryUrl,
-      settings.clean,
-      settings.ref
-    )
-  }
+  let authHelper: gitAuthHelper.IGitAuthHelper | null = null
```
TingluoHuang, line 39: what is the `| null` for?

src/git-source-provider.ts
```diff
@@ -36,68 +36,77 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
-    core.warning(
-      `Unable to turn off git automatic garbage collection. The git fetch operation may trigger garbage collection and cause a delay.`
-    )
-  }
-  core.endGroup()
-
-  const authHelper = gitAuthHelper.createAuthHelper(git, settings)
-  try {
+    // If we didn't initialize it above, do it now
+    if (!authHelper) {
```
TingluoHuang, line 107: will this ever be true? `if (!git) { return }` on line 81

src/git-source-provider.ts
thboop: Creates a nullable type, so the value can be an object of IGitAuthHelper, or null

src/git-auth-helper.ts
thboop: Yeah, the fourth parameter is add

```
  config(
    configKey: string,
    configValue: string,
    globalConfig?: boolean,
    add?: boolean
```

src/git-auth-helper.ts
thboop: removed!

src/git-source-provider.ts
```diff
@@ -170,34 +179,26 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
-        )
-        await git.submoduleForeach(
-          'git config --local gc.auto 0',
-          settings.nestedSubmodules
-        )
-        core.endGroup()
+      // Checkout submodules
+      core.startGroup('Fetching submodules')
+      await git.submoduleSync(settings.nestedSubmodules)
+      await git.submoduleUpdate(settings.fetchDepth, settings.nestedSubmodules)
```
thboop, line 190: linter did this whitespace change

src/git-auth-helper.ts
```diff
@@ -110,13 +115,33 @@ class GitAuthHelper {
...
       )
-      this.git.setEnvironmentVariable('HOME', this.temporaryHomePath)
+      .catch(error => {
+        core.info(`Failed to initialize safe directory with error: ${error}`)
+      })
+    return newGitConfigPath
+  }
 
+  async configureGlobalAuth(): Promise<void> {
+    // noops if already set, just returns the path
```
TingluoHuang, line 142: this comment might be wrong?

src/git-auth-helper.ts
```diff
@@ -110,13 +115,33 @@ class GitAuthHelper {
       await fs.promises.writeFile(newGitConfigPath, '')
     }
 
-    try {
-      // Override HOME
```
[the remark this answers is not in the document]
thboop: I didn't think it was needed given the next two lines, but I can add it back!

dist/index.js
```diff
@@ -7415,27 +7445,21 @@ function getSource(settings) {
...
-                    yield git.submoduleForeach('git config --local gc.auto 0', settings.nestedSubmodules);
+                // Temporarily override global config
+                core.startGroup('Setting up auth for fetching submodules');
+                yield authHelper.configureGlobalAuth();
+                core.endGroup();
+                // Checkout submodules
+                core.startGroup('Fetching submodules');
+                yield git.submoduleSync(settings.nestedSubmodules);
+                yield git.submoduleUpdate(settings.fetchDepth, settings.nestedSubmodules);
+                yield git.submoduleForeach('git config --local gc.auto 0', settings.nestedSubmodules);
```
JasonGross, line 7456: I think there should be another `git.submoduleForeach` line after this one that adds submodules to the safe path.  Reported as https://github.com/actions/checkout/issues/768

liskin:
> This does carry some limitations, it doesn't persist this configuration for the duration of the job mainly. So if your job pushes to git or something after checkout, that will continue to fail. We need to figure out how to address this at an ecosystem level, outside of the checkout action.

This doesn't just affect stuff like pushing to git, it breaks even relatively common things like "git status", "git describe", basically any git command at all. Therefore anyone who uses that sort of information in their jobs will see breakage. In my case, https://github.com/pypa/setuptools_scm no longer works. Is there an issue tracking the "addressing this at an ecosystem level" which I can watch and get notified when it's resolved? Thanks!

PR #776, opened 2022-04-21
Add set-safe-directory input to allow customers to take control. (#770)

TingluoHuang:
* Add set-safe-directory input to allow customers to take control.


thboop (approved):
lgtm

Issue #630, opened 2021-10-31
Please add support for checkouting external repositories


mwestphal:
External repositories are not supported, eg:

    - name: Checkout OCCT
      uses: actions/checkout@v2
      with:
        repository: https://git.dev.opencascade.org/repos/occt.git 
        submodules: 'true'
        path: './dependencies/occt'
        ref: V7_5_1

Does not work at all

mwestphal:
mentioned multiple times here: https://github.com/actions/checkout/issues/24

Merinorus:
Hello,

I have a similar problem: my GitHub instance is self-hosted (github.mycompany.com) and I try to checkout a repo on GitHub (github.com).

I could not find any solution in discussion #24. They talk about mixing public and private repo but they are all hosted on github.com.

The only way I could bypass this was by doing a bash command `git checkout`.

PR #830, opened 2022-06-15
Add Config Options to Override github Host

GABeech:
This is possible solution to #630 it certainly allows you to override between github instances, but would need some investigation to see if putting a non github host in would work properly.

All this does is simply add an optional input called `setHost` that will be returned instead of the github urls if defined.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -71,6 +71,13 @@ inputs:
   set-safe-directory:
     description: Add repository path as safe.directory for Git global config by running `git config --global --add safe.directory <path>`
     default: true
+  
+    setHost:
+      description: >
+        override the github host that is set automatically to the github instance the action was called from. 
+        This is useful if you need to clone from an instance that is not the one your action is running from.  
+        For example, you need to clone a cloud hosted repo from an action run by an on prem instance.
+      required: false
 runs:
   using: node16
   main: dist/index.js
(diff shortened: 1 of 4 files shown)


action.yml
```diff
@@ -71,6 +71,13 @@ inputs:
   set-safe-directory:
     description: Add repository path as safe.directory for Git global config by running `git config --global --add safe.directory <path>`
     default: true
+  
+    setHost:
```
jk2K, line 75: setHost -> set-host

src/git-auth-helper.ts
```diff
@@ -52,7 +52,7 @@ class GitAuthHelper {
     this.settings = gitSourceSettings || (({} as unknown) as IGitSourceSettings)
 
     // Token auth header
-    const serverUrl = urlHelper.getServerUrl()
+    const serverUrl = urlHelper.getServerUrl(gitSourceSettings?.setHost)
```
jk2K, line 55: change to
```diff
- const serverUrl = urlHelper.getServerUrl()
+ const serverUrl = urlHelper.getServerUrl(this.settings.setHost)
```

src/url-helper.ts
```diff
@@ -1,14 +1,15 @@
 import * as assert from 'assert'
 import {IGitSourceSettings} from './git-source-settings'
 import {URL} from 'url'
+import { settings } from 'cluster'
```
jk2K, line 4: this is useless

jk2K:
`input-helper.ts` file should be change

```diff
  // Set safe.directory in git global config.
  result.setSafeDirectory =
    (core.getInput('set-safe-directory') || 'true').toUpperCase() === 'TRUE'

+  // set host
+  result.setHost = (core.getInput('set-host') || '')
+  return result
```

Issue #563, opened 2021-08-06
Is it possible to checkout to enterprise github repo from non enterprise public repo in github actions?


LakshmiRavali:
I want to checkout to enterprise github repo from non-enterprise public repo. These are my actions:

```
jobs:
  checkout:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [10.x]
    steps:
    - name: Checkout private tools
      uses: actions/checkout@v2
      with:
        repository: enterprise/repo
        token: ${{ secrets.ENTERPRISE_TOKEN }} 
        path: main
```
but as I saw it will clone https://github.com/enterprise/repo, but my enterprise account is on a different domain. Is it possible to checkout to the enterprise account?

LakshmiRavali:
When I am checking the code, there is a [environment variable](https://github.com/actions/checkout/blob/25a956c84d5dd820d28caab9f86b8d183aeeff3d/src/url-helper.ts#L25) for server url. I am trying to set it like this:
```
jobs:
  checkout:
    runs-on: ubuntu-latest
    env:
        GITHUB_SERVER_URL: {{ENTERPRISE_URL}}
        GITHUB_API_URL: {{ENTERPRISE_URL}}
    strategy:
      matrix:
        node-version: [10.x]
    steps:
    - name: Checkout private tools
      uses: actions/checkout@v2
      with:
        repository: enterprise/repo
        token: ${{ secrets.ENTERPRISE_TOKEN }} 
        path: main
```
but still, I am getting the error 
```
Retrieving the default branch name
  Error: Bad credentials
```
Can anyone let me know how to use the environment variable to set the server URL?

lancetarn:
I would like to do this as well. It seems impossible with this action. Overriding environment variables namespaced with `GITHUB_` is explicitly forbidden as described [here](https://docs.github.com/en/actions/learn-github-actions/environment-variables#naming-conventions-for-environment-variables)

Makes me wonder if things would break if this was turned into an `input` that the url-helper respected.

PR #844, opened 2022-07-03
fix: octokit to use the correct enterprise url

seehait:
fixes https://github.com/actions/checkout/issues/563

This is just a quick fix. We should define a new input to properly specify the custom server URL.

I found out that even when we override the `GITHUB_SERVER_URL` environment variable, the `getDefaultBranch` initializes a new Octokit client without the `baseUrl` value regarding `GITHUB_SERVER_URL`.
Thus, the checkout action fails when we try to checkout to a private Github Enterprise Server repository with a misguided error messages.

I don't know the precise reasons why we haven't allowing any action users to customize the server URL via input though.
So, any comments are welcome.


seehait:
Need to setup a proper input anyway

PR #922, opened 2022-09-10
Inject GitHub host to be able to clone from another GitHub instance

peter-murray:
These changes allow for the base URL of another GitHub instance to be provided to the action so that a workflow running on a GHES server can clone a repository from another GHES server or github.com.

These changes maintain the existing default behaviour whilst allowing for a user to inject a URL to alternate GitHub instance and have been validated using an actions runner with and without a git command line client (i.e. git command line clone and HTTP cloning) have been tested.


__test__/git-auth-helper.test.ts
```diff
@@ -778,7 +778,8 @@ async function setup(testName: string): Promise<void> {
     sshKnownHosts: '',
     sshStrict: true,
     workflowOrganizationId: 123456,
-    setSafeDirectory: true
+    setSafeDirectory: true,
+    githubServerUrl: undefined
```
TingluoHuang, line 782: can we have a test with `githubServerUrl` set to some URL?

src/octokit-helper.ts
```diff
@@ -0,0 +1,23 @@
+import * as github from '@actions/github'
+import {Octokit} from '@octokit/rest'
+import {getServerApiUrl} from './url-helper'
+
+// Centralize all Octokit references by re-exporting
+export {Octokit} from '@octokit/rest'
+
+export type OctokitOptions = {
+  baseUrl?: string
+  userAgent?: string
```
TingluoHuang, line 10: anything set this `userAgent`?

peter-murray: Yes the `ref-helper.ts` injects this currently, https://github.com/peter-murray/checkout/blob/server-url-injection/src/ref-helper.ts#L250

__test__/git-auth-helper.test.ts
peter-murray: Added tests to the branch and refactored the one that was there to use common code for that test

src/octokit-helper.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function getOctokit(authToken: string, opts: OctokitOptions) {
+  const options: Octokit.Options = {
+    baseUrl: getServerApiUrl(opts.baseUrl)
```
TingluoHuang, line 15: can we also conditionally set the baseUrl when the `opts.baseUrl` is provided?

peter-murray: That is technically what the function that is being called is doing, as it replicates the octokit default behaviour of a fallback to github.com/api, https://github.com/actions/checkout/pull/922/files#diff-f53c8d0b0ecfe6c87df94f6060fbc83aa0ec2c24f279c8025be59febb6a68dc0R22

it is possible to set this conditionally, but it is less obvious that this is the behaviour as the default in octokit is a separate library and buried deeper in the code base.

Link- (changes requested):
@peter-murray can you please run `npm run format` and push the changes?

Otherwise, I don't see any glaring issues with this change.

src/octokit-helper.ts
TingluoHuang: Understand, my view was for safe rollout and minimizing any risk.
We were relying on `@actions/github` to populate the default baseUrl, in case there is any change in how to populate the baseUrl in the default case, upgrading `@actions/github` to the latest will handle it for us.

peter-murray: I can indeed understand that. This is not the only place where you have defaults like this.
The way that I implemented this was to consolidate the already default fallback nature in a way that is consistent for the api and host, as the code base has an implicit fallback on api.github.com vs the GHES equivalent and one on a GHES server base url or https://github.com. To properly connect these decisions together as one is defaulted in this code base and the other is hurried layers deep in the octokit wrapper, that is why I did it this way.

Sure it can be done to utilize the fallback, but then you have one assumption here in this action for the git commands and another burried deep inside octokit that operates on a default, that in all honesty is never used as the environment URLs are already populated and used as fallbacks before you get that empty behaviour in actions in the first place.

I am happy to refactor this again if it is perceived to be a blocking factor in this change, but I think in all honesty it removes clarity and will make the changes for injection potentially more messy than what we have, as well as leading to a potential inadvertent divergent change.

Link- (changes requested):
Minor feedback, otherwise LGTM

src/octokit-helper.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// Centralize all Octokit references by re-exporting
+export {Octokit} from '@octokit/rest'
+
+export type OctokitOptions = {
+  baseUrl?: string
+  userAgent?: string
+}
+
+export function getOctokit(authToken: string, opts: OctokitOptions) {
```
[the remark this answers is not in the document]
peter-murray: Sure, I just went with `-helper` which was consistent with the other files. I did also move the `@octokit/rest` import and export into this file too so that all octokit references are made from this file keeping it all together.

I have renamed it to `-provider`

README.md
```diff
@@ -97,6 +97,11 @@ When Git 2.18 or higher is not in your PATH, falls back to the REST API to downl
     # config --global --add safe.directory <path>`
     # Default: true
     set-safe-directory: ''
+
+    # The base URL for the GitHub instance that you are trying to clone from, will use
+    # environment defaults to fetch from the same instance that the workflow is
+    # running from unless specified.
```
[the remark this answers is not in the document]
peter-murray: Added examples to the input variable so that will update the README.md

Link- (approved):

TingluoHuang (approved):

Issue #795, opened 2022-05-15
Add option to checkout via REST API


ddzz:
Currently, if git is not installed, a repo is downloaded via the GitHub REST API. An option should be added to download a repository via REST even when git is installed.

yogurtearl:
Downloads via REST API can be 2 to 3x faster in some cases, so would be nice to be able to explicitly choose that option.

PR #936, opened 2022-10-01
Adds force-api-download option

collinmcneese:
Signed-off-by: CollinM <collinmcneese@github.com>

Adds `force-api-download` option to allow for downloading refs using REST API instead of installed `git` exec in instances where it would be more performant to do so related to issue #795


toast-gear:
would this not be more extensible and slightly cleaner if instead the new input was to define a default method rather than specifically forcing using the api e.g.

```yaml
# accepts  `git-client` or 'api`
default-method: 'git-client`
```

PR #946, opened 2022-10-05
Default for persist-credentials should be false

sebbASF:
Does not seem right to expose the credentials by default.


randydwhitephotography (approved):
signed-off-by Randy D White randydwhite@randydwhite.com

PR #963, opened 2022-10-14
wrap the command for submoduleForeach in quotes

jokreliable:
This is a poor attempt at fixing the function to work as expected when `command` is a pipeline of some sort.

This may be a fix for #354 and others.


jokreliable:
I think it might be better to fix the commands themselves rather than try to wrap things in quote every time... abandoning this for now.

PR #964, opened 2022-10-14
wrap pipeline commands for submoduleForeach in quotes

jokreliable:
Update the commands passed to submoduleForeach to make sure the expected behavior is happening.
Without the quotes, only the first part of those pipelines are handled by git-submodule foreach, and the rest of the pipeline is handled outside of the loop, which is likely not the intended behavior.


jokreliable:
This should address #354 and a few other issues related to submodule(s).

src/git-auth-helper.ts
```diff
@@ -365,7 +366,8 @@ class GitAuthHelper {
 
     const pattern = regexpHelper.escape(configKey)
     await this.git.submoduleForeach(
-      `git config --local --name-only --get-regexp '${pattern}' && git config --local --unset-all '${configKey}' || :`,
+      // wrap the pipeline in quotes to make sure it's handled properly by submoduleForeach, rather than just the first part of the pipeline
+      `"git config --local --name-only --get-regexp '${pattern}' && git config --local --unset-all '${configKey}' || : "`,
```
jsoref, line 370: Personally, I'd drop this space...
```suggestion
      `"git config --local --name-only --get-regexp '${pattern}' && git config --local --unset-all '${configKey}' || :"`,
```

jsoref (approved):
This makes sense.

jsoref:
@cory-miller?

cory-miller:
I sent a [PR](https://github.com/actions/checkout/pull/1002) to fix an unrelated CI issue. In the meantime, could you please update the JS bundle with `npm run build`?

jokreliable:
Updated! (and thank you for the pointers -- never used typescript before...)

jokreliable:
yikes - that's a lot of failures.. not sure I can help much with making the CI pass... They all seem to point to some sort of dependency resolver for npm (and i'm guessing it's part of the package-lock.json that i included as part of my commit after i updated the dist file... any pointer/docs I can go look at to get educated on this?

cory-miller:
Sorry for the delay and bad info. I assumed this Action worked similarly to others owned by my team, but that doesn't seem to be the case. Instead of `npm install`, you can use `npm ci`. These commands should fix the deps problems in CI.

```
git checkout origin/main -- package-lock.json && \
  npm ci && npm run format && npm run build
```

I did the same in https://github.com/actions/checkout/pull/1010 and the CI now fails due to submodule command problems: https://github.com/actions/checkout/actions/runs/3463800067/jobs/5784532604

The failing CI is https://github.com/actions/checkout/blob/5c3ccc22eb2c950a0fa5bc7c47190d8e3f7e681a/.github/workflows/test.yml#L96-L102

The exact error copied below in case you can't see.

```
  /usr/bin/git submodule foreach --recursive "git config --local --name-only --get-regexp 'url\.https\:\/\/github\.com\/\.insteadOf' && git config --local --unset-all 'url.https://github.com/.insteadOf' || :"
  Entering 'submodule-level-1'
  path='submodule-level-1'; "git config --local --name-only --get-regexp 'url\.https\:\/\/github\.com\/\.insteadOf' && git config --local --unset-all 'url.https://github.com/.insteadOf' || :": 1: git config --local --name-only --get-regexp 'url\.https\:\/\/github\.com\/\.insteadOf' && git config --local --unset-all 'url.https://github.com/.insteadOf' || :: not found
```

Not having worked in actions/checkout before, I am still trying to figure out why this is broken. Let me know if you have insight. I'm also reaching out to other on my team that have worked in this action before.

jokreliable:
I expect that git, instead of just running the code in there as a shell command (the part i wrapped quotes around), it is trying to use the whole thing (including spaces and all that) as $0 in exec... so not quite as simple as that. There might need to be some trickery required, or some extensive rework of the code to make it work... I'll have a look and will try a couple of different things to see what works...

jokreliable:
done! (I had to sync specific branches into my repo, but the CI cleared all green across the board... (see https://github.com/jokreliable/checkout/actions/runs/3475087417 for output) -- will update this branch to match...

jokreliable:
well, outside of the comment present in this branch, it matches... (i think)

Issue #494, opened 2021-05-05
Suppress hint: You are in 'detached HEAD' state.


jsoref:
Downstream consumers of actions shouldn't see noise that git is checking out a detached head. When `actions/checkout` is used, it will be given something specific to check out, and if that results in a detached head, that's clearly intentional...

```yaml
name: Spell checking
on:
  pull_request_target:

jobs:
  check-spelling:
    name: Spell checking
    runs-on: ubuntu-latest
    steps:
    - name: checkout-merge
      if: "contains(github.event_name, 'pull_request')"
      uses: actions/checkout@v2.0.0
      with:
        ref: refs/pull/${{github.event.pull_request.number}}/merge
```

```.
Run actions/checkout@v2.0.0
  with:
    ref: refs/pull/1/merge
    repository: check-spelling/sticky-pull-request-comment
    token: ***
    persist-credentials: true
    clean: true
    fetch-depth: 1
    lfs: false
Syncing repository: check-spelling/sticky-pull-request-comment
Working directory is '/home/runner/work/sticky-pull-request-comment/sticky-pull-request-comment'
...
From https://github.com/check-spelling/sticky-pull-request-comment
 * [new ref]         refs/pull/1/merge -> pull/1/merge
/usr/bin/git checkout --progress --force refs/remotes/pull/1/merge
Note: switching to 'refs/remotes/pull/1/merge'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 4ea2184 Merge b9cf99a584fe42271c3aaee8ad2337ed41fe71b5 into 49d94c2e2fd173b2af1fedaabb95dbd37c0e80e4
/usr/bin/git log -1
commit 4ea21846857440dff5c72b5cd423d1f8d56400c5
Author: Josh Soref <jsoref@users.noreply.github.com>
Date:   Wed May 5 08:47:45 2021 -0400

    Merge b9cf99a584fe42271c3aaee8ad2337ed41fe71b5 into 49d94c2e2fd173b2af1fedaabb95dbd37c0e80e4
```

vickyavillaavila84-dotcom:
[...]

PR #981, opened 2022-10-24
Suppress "detached head" advice during checkout

jeffwidman:
Checking out certain `ref` values will result in a warning about a detached `HEAD`:
```
You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false
```

However, this warning isn't useful in a CI environment... so suppress it.

I realize on the original bug report that one user mentioned this warning highlighted a bug in his actions flow, but I consider that a super rare / happy accident. 99% of use cases will be _intentionally_ checking out a specific ref where the detached head state is inevitable, so the warning is pure noise.

Passing the config this way sets it _only_ for this command. Note that it must be set [_before_ calling `checkout`](https://stackoverflow.com/a/72588008/770425).

Resolve: https://github.com/actions/checkout/issues/494

### DIFF
--- a/src/git-command-manager.ts
+++ b/src/git-command-manager.ts
@@ -123,7 +123,7 @@ class GitCommandManager {
   }
 
   async checkout(ref: string, startPoint: string): Promise<void> {
-    const args = ['checkout', '--progress', '--force']
+    const args = ['-c advice.detachedHead=false', 'checkout', '--progress', '--force']
     if (startPoint) {
       args.push('-B', ref, startPoint)
     } else {


jeffwidman:
I'm not very familiar with typescript, so unclear if I need to also compile this to the javascript `dst` code. 

Happy to do that if you let me know what I need to do.

Frieder-AirLST:
I have an issue in my flow right now. the message you want to suppress brought me here... please do not disable it by default.

PR #1025, opened 2022-12-01
Update @actions/io to pickup bug fixes

cory-miller:
Also updates `package-lock.json` which seems to have been built with an older version of `npm`.


vmjoseph (approved):

package.json
```diff
@@ -31,19 +31,22 @@
     "@actions/core": "^1.10.0",
     "@actions/exec": "^1.0.1",
     "@actions/github": "^2.2.0",
-    "@actions/io": "^1.0.1",
+    "@actions/io": "^1.1.2",
     "@actions/tool-cache": "^1.1.2",
+    "minimatch": "^5.1.1",
+    "npm": "^8.19.2",
```
johnsudol, line 37: I'm curious if we need `npm` as a dependency here? could we use an `engine` object inside of the file instead? something like this
```json
"engines": {
    "node": ">=12",
    "npm": ">=6"
  },
```

cory-miller: @vmjoseph could you help with this? `npm` was added as a dep in one of your commits and not sure the reason.

PR #833, opened 2022-06-20
Fix: Don't remove auth if submodules are off

sevenc-nanashi:
Fix the action tries to remove auth even the submodules are disabled.
Fixes #610.

PR #1051, opened 2022-12-17
removeGitConfig should follow nested submodule setting for not

tjjh89017:
`removeGitConfig` didn't follow the `nestedSubmodule`

```
Error: fatal: No url found for submodule path '***********' in .gitmodules
Error: fatal: run_command returned non-zero status while recursing in the nested submodules of buildroot
```


tomasbw:
Does it address the same issue as https://github.com/actions/checkout/pull/833 ?

tjjh89017:
> Does it address the same issue as #833 ?

I think these are two issues.
 #833 will not fix my issue.
Because I have submodule, but I want to disable nested submodule. (Submodule=on, nested submodule=off)
#833 only consider submodule but not nested submodule.

PR #579, opened 2021-09-08
Add option to fetch tags even if fetch-depth > 0

RobertWieczoreck:
Hi there!

This PR adds an optional `fetch-tags` option that can be used to override the default behavior of adding the `--no-tags` arg to the generated `git fetch` command if `fetch-depth` is configured and not `0`.

This way, users can use fetch-depth while still getting the tags inside this range, which can e.g.  be important when using automatic GIT versioning plugins like [jgitver](https://github.com/jgitver/jgitver).

Tested using a workflow on a separate branch that verifies function as well as backwards compatibility: See [Run 5432963197](https://github.com/RobertWieczoreck/checkout/actions/runs/5432963197/jobs/9880354020).

All Checks are green.

Also see [PR 2](https://github.com/RobertWieczoreck/checkout/pull/2) in my fork (description is mostly copied from there).

Fixes #448 

**EDIT:** Adapted to the recent changes regarding sparse checkout and added Jest tests.

PR #1067, opened 2023-01-05
Support fetching without the --progress option

### LINKED ISSUE #894: Run git fetch with --quiet by default
https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt--q

As has been noted before in the prematurely closed #489 and #18 the fetch progress does not add any value and just clutters the logs, esp. the raw logs.
The folding in the UI does not fix the issue for the raw logs.
And in general it's a waste of bandwidth and everything.

simonbaird:
If this is merged, setting the `progress` option to false in the `with` section of the workflow step will cause git fetch to run without the `--progress` flag.

Note that the default value of this new option is `true`, which matches the current behavior, so this PR introduces no behavior change unless the new option is explicitly present and set to false.

The motivation is to be able to suppress the noisy progress status output which adds many hundreds of `remote: Counting objects: 85% (386/453)` and similar lines in the workflow log.

This should be sufficient to resolve #894 and its older friends, though the solution is different to the one proposed there because it doesn't use the `--quiet` flag. IIUC git doesn't show the progress status by default when running on GitHub since the output is not a terminal. Adding the `--progress` flag is what forces the progress output to be shown, so that's why removing it is all that's needed.

Adding the `--quiet` flag doesn't make a lot of difference once the `--progress` flag is removed, and actually I think using `--quiet` would suppress some other more useful output that would be better left visible.

### DIFF
--- a/__test__/git-command-manager.test.ts
+++ b/__test__/git-command-manager.test.ts
@@ -135,7 +135,6 @@ describe('Test fetchDepth and fetchTags options', () => {
         'protocol.version=2',
         'fetch',
         '--prune',
-        '--progress',
         '--no-recurse-submodules',
         '--filter=filterValue',
         'origin',
@@ -174,7 +173,6 @@ describe('Test fetchDepth and fetchTags options', () => {
         'fetch',
         '--no-tags',
         '--prune',
-        '--progress',
         '--no-recurse-submodules',
         '--filter=filterValue',
         'origin',
@@ -213,7 +211,6 @@ describe('Test fetchDepth and fetchTags options', () => {
         'fetch',
         '--no-tags',
         '--prune',
-        '--progress',
         '--no-recurse-submodules',
         '--filter=filterValue',
         '--depth=1',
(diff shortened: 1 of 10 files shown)


simonbaird:
I'm using it [here](https://github.com/simonbaird/tiddlyhost/actions/runs/3844090549/jobs/6546904868#step:2:50) and it seems to work well.

simonbaird:
[This workflow run](https://github.com/simonbaird/tiddlyhost/actions/runs/3844120635/jobs/6546957340#step:2:50) shows it working with the `--progess output` which is the default.

simonbaird:
@TingluoHuang would this be something you'd consider taking a look at? I'm looking at your comments from 2019 on #18.

yeikel:
What is blocking this change? 

I am looking forward to seeing this released as well

simonbaird:
Rebased on upstream/main, (currently f095bcc).

simonbaird:
Passes formatting and lint checks for me locally:

```
for t in build format-check lint; do npm run $t; done

> checkout@3.5.2 build
> tsc && ncc build && node lib/misc/generate-docs.js

ncc: Version 0.36.1
ncc: Compiling file index.js into CJS
877kB  dist/index.js
877kB  [1343ms] - ncc 0.36.1

> checkout@3.5.2 format-check
> prettier --check '**/*.ts'

Checking formatting...
All matched files use Prettier code style!

> checkout@3.5.2 lint
> eslint src/**/*.ts
```

simonbaird:
I reworded the PR description and added some addition explanations.

simonbaird:
Rebased on 96f5310.

simonbaird:
```
$ for t in build format-check lint; do npm run $t; done

> checkout@3.5.3 build
> tsc && ncc build && node lib/misc/generate-docs.js

ncc: Version 0.36.1
ncc: Compiling file index.js into CJS
879kB  dist/index.js
879kB  [1112ms] - ncc 0.36.1

> checkout@3.5.3 format-check
> prettier --check '**/*.ts'

Checking formatting...
All matched files use Prettier code style!

> checkout@3.5.3 lint
> eslint src/**/*.ts
```

slang25:
@luketomlinson are you able to comment, there are a lot of people wanting this.

yeikel:
Well, currently there are conflicts. It would help if we can resolve these first. Otherwise it'll take more time before this is looked at again 

@simonbaird Could you please work on that?

simonbaird:
Rebased on upstream/main (currently at 97a652b80035363df47baee5031ec8670b8878ac). Now includes some test coverage :sunglasses: .

__test__/git-command-manager.test.ts
```diff
@@ -263,4 +259,120 @@ describe('Test fetchDepth and fetchTags options', () => {
...
+      doSparseCheckout
+    )
+    const refSpec = ['refspec1', 'refspec2']
+    const options = {
+      filter: 'filterValue',
+      fetchDepth: 42,
+      showProgress: true
+    }
+
+    await git.fetch(refSpec, options)
```
simonbaird, line 319: It might be nice to merge options onto the default options here so the assertions become more like real world examples. Would be a separate commit though.

__test__/git-command-manager.test.ts
```diff
@@ -263,4 +259,120 @@ describe('Test fetchDepth and fetchTags options', () => {
...
+        '--no-recurse-submodules',
+        '--progress',
+        '--filter=filterValue',
+        'origin',
+        'refspec1',
+        'refspec2'
+      ],
+      expect.any(Object)
+    )
+  })
```
simonbaird, line 377: I'd be happy to remove some of the test coverage if it's considered too much. I was trying to follow the example set in #579 .

vanZeben (approved):
Thanks for putting this together @simonbaird, especially with the comprehensive test coverage! 🎉  I'll cut a release for this later today

vanZeben:
I lied, checking internally, we're planning on cutting a new Major version release next week to include the [recent node20 bump](https://github.com/actions/checkout/pull/1436), that major release will include this too, so look forward to `v4.0.0` for this change!

yeikel:
> I lied, checking internally, we're planning on cutting a new Major version release next week to include the [recent node20 bump](https://github.com/actions/checkout/pull/1436), that major release will include this too, so look forward to `v4.0.0` for this change!

Can this be backported to 3.x as well?

I am running this in an environment where the node20 runners are not available yet

Gakk:
@simonbaird Setting `show-progress: false` gets rid of the initial `Counting objects`, `Compressing objects`, `Receiving objects` and `Resolving deltas` under the step `Enumerating objects` 👍

But the step `Checking out the ref` still runs with `--progress` adding a lot of log lines for `Updating files:`

image

I get this when using `runs-on: [windows-latest]` and:

```yml
    - name: Checkout
      uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
      # https://github.com/actions/checkout/releases
      with:
        show-progress: false
        fetch-depth: 0
```

And also with `runs-on: [windows-2019]` and:

```yml
    - name: Checkout
      uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
      with:
        show-progress: false
```

simonbaird:
You're 100% right, the changes from the `fetch` method need to be also done for the `checkout` method. I'll take a look. But let's file a new issue.

simonbaird:
> You're 100% right, the changes from the `fetch` method need to be also done for the `checkout` method. I'll take a look. But let's file a new issue.

See #1455.

AngelicaOvesdotter (commented):
🤩

PR #1137, opened 2023-01-22
Workflow: update to node 18.x

GiddyGoatGaming:
Updates node to 18.x from 16.x in the workflow folder

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -24,10 +24,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set Node.js 16.x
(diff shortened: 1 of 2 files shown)


Jdzimmo (approved):
5*****

Moser-ss:
Just for curiosity why test with node 18.x if the actions are executed with node 16.16, defined [here](https://github.com/actions/runner/blob/542e8a3c9832239180fd65cde5154e0e059e28ec/src/Misc/externals.sh#L8)

GiddyGoatGaming:
Oh shoot, I did not see that when looking through the repo. Il close this now as it doesn’t seem fit to bump node

Issue #1205, opened 2023-03-09
id diff missing on step workflow check-dist


alexquitiaquez:
Hi,
I found that this condition is not working `steps.diff.conclusion == 'failure'` cause there is no id for previous step that does the comparison
https://github.com/actions/checkout/blob/ac593985615ec2ede58e132d2e21d2b1cbd6127c/.github/workflows/check-dist.yml#L48

All you need is to add the `id: diff` between these two lines
https://github.com/actions/checkout/blob/ac593985615ec2ede58e132d2e21d2b1cbd6127c/.github/workflows/check-dist.yml#L38-L39

Without that, the action upload-artifact will never upload the artifact dist/

thanks

alexquitiaquez:
I opened a PR https://github.com/actions/checkout/pull/1206

PR #1206, opened 2023-03-09
Adding id: diff to step doing comparison and update of upload-artifac…

alexquitiaquez:
…t action to v3 (node 16))

Fix to issues/1205


alexquitiaquez:
Fixing #1205

Issue #1186, opened 2023-02-23
Checkout suddenly much slower on windows environment


yeu-degroot:
> Moved from <https://github.com/actions/runner-images/issues/7180>

### Description

Since a few days, the duration of the `actions/checkout@v3` steps on `windows-2019` have dramatically increased.
This behavior is seen on all my repos (all private). Below a table showing an example of before/after.

| Checkout step | Feb 17th | Feb 21st |
| -- | -- | -- |
| 1 | `13s` | `1m35s` |
| 2 | `8s` | `47s` |

The result is a huge increase in build (and billable) time.

![image](https://user-images.githubusercontent.com/52652923/220855259-79a6e246-acd8-40b8-a9f3-7865cda67e34.png)

![image](https://user-images.githubusercontent.com/52652923/220855227-566903a4-42ad-4de9-8414-402171c4dcfe.png)

The [github status](https://www.githubstatus.com/history) page does show some issues around this time frame, but these were all solved:

![image](https://user-images.githubusercontent.com/52652923/220855851-9c8dac29-19c5-43ca-9835-06b043285087.png)

### Platforms affected

- [ ] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [X] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

```
Current runner version: '2.301.1'
Operating System
  Microsoft Windows Server 2019
  10.0.17763
```

Private repo

### Is it regression?

Yes, sorry private repos

### Expected behavior

The build times should be fairly constant. 

### Actual behavior

Build times explode. Burning down our build minutes too fast.

### Repro steps

Compare build times on any windows environment from before Feb 18th with today.

Tyrrrz:
I have observed this too. Example job: https://github.com/Tyrrrz/CliWrap/actions/runs/4271628026

- **Windows** – checkout takes **4m9s**:

![image](https://user-images.githubusercontent.com/1935960/222046611-680120d3-b5b3-4dd7-bc15-5f46ec8239ad.png)

- **Ubuntu** – checkout takes **<1s**:

![image](https://user-images.githubusercontent.com/1935960/222046665-029b86cc-968d-41cd-b0d8-711963d9cbd0.png)

[...]

PR #1246, opened 2023-03-26
Improve checkout performance on Windows runners by upgrading @actions/github dependency

BrettDong:
Re: https://github.com/actions/checkout/issues/1186

@dscho discovered that the checkout action could stall for a considerable amount of time on Windows runners waiting for PowerShell invocations made from 'windows-release' npm package to complete.

Then I studied the dependency chain to figure out where 'windows-release' was imported:

```
'@actions/checkout'
  <- '@actions/github'@2.2.0
    <- '@octokit/endpoint'@6.0.1
    <- '@octokit/graphql'@4.3.1
    <- '@octokit/request'@5.4.2
    <- '@octokit/rest'@16.43.1
      <- 'universal-user-agent'@4.0.1
        <- 'os-name'@3.1.0
          <- 'windows-release'@3.1.0
```

'universal-user-agent' package dropped its dependency on 'os-name' in v6 (https://github.com/gr2m/universal-user-agent/releases/tag/v6.0.0).

'@actions/github' v3 removed dependency on '@octokit/rest'@16.43.1 and allows users to move away from the old 'universal-user-agent' v4. (https://github.com/actions/toolkit/pull/453)

This pull request attempts to update the version of '@actions/github' used in the checkout action to avoid importing 'windows-release'.

Based on testing in my own repositories, I can see an improvement in reduced wait time between entering the checkout action and git actually starts to do useful work.


dscho (approved):
I confirm that this fixes the abysmal slowness on Windows that was reported in #1186 and that I was able to reproduce.

Well done, @BrettDong!

briantist:
Eagerly awaiting a fix for this; for those of us paying for runners it's not just a convenience, we end up paying significantly more because the checkout step takes the majority of the runner's time, and with a workflow that has a lot of parallel runs, this adds up really quickly.

briantist:
I've been using this and it was faster, but I just had a run with it that took 1m41s 😖

src/github-api-helper.ts
```diff
@@ -134,7 +134,7 @@ async function downloadArchive(
     repo: repo,
     ref: commit || ref
   })
-  if (response.status != 302) {
+  if (response.status != 200) {
```
dscho, line 137: could we accept both 200 and 302? And: could the commit message please provide context for this change?

BrettDong: This was edited on mobile phone to see if it can make CI pass, but unfortunately seems running CI requires manual approval. I'll do a proper amend commit on my computer once I have time.

BrettDong:
Hi @thboop , could you please re-approve the CI checks? Thanks!

Spiritaine (approved):

src/github-api-helper.ts
```diff
@@ -125,19 +125,20 @@ async function downloadArchive(
...
+    : octokit.rest.repos.downloadTarballArchive
+  const response = await download({
     owner: owner,
     repo: repo,
-    archive_format: IS_WINDOWS ? 'zipball' : 'tarball',
     ref: commit || ref
-  }
-  const response = await octokit.repos.getArchiveLink(params)
+  })
   if (response.status != 200) {
```
BrettDong, line 137: OK I need help on how to properly detect error from the status code.
```js
if (response.status != 200) {
```
causes a compile error saying the comparison is always false.

BrettDong:
```js
if (response.status != 302) {
```
does not work either because its value is 200 not 302 in a successful request.

fhammerl: At a glance, it looks like a 302 response code is guaranteed at this point, implying that `octokit.rest.repos.downloadZip|TarballArchive` would throw in case it fails.

So we don't need to explicitly check the response code here anymore IMO, `download` will throw and it'll propagate up.

fhammerl:
Hi @BrettDong, thanks for this PR! Updating to actions/github@v5 is both something we want to do and it sidesteps the whole issue by eliminating our dependency on `windows-release` by removing `universal-user-agent` all together, so thanks for taking that approach :) 

A few notes / questions:

1. Do you have a minimal repro that only requires one of `windows-release`? E.g. I tried to run `console.log(windowsRelease())` on a larger runner but it did not cause the same stalling behaviour. Running `(Get-CimInstance -ClassName Win32_OperatingSystem).caption` directly in powershell is also instantaneous
2. I could only reproduce this issue on larger runners, not on `windows-latest` or `windows-2019`. Do you experience it there too?

BrettDong:
The delay from `windows-release` is only part of the reason checkout is slow on Windows. This pull request solves specifically this single issue. In majority of cases the delay of `windows-release` is of single digit seconds; in rarer cases it takes longer time. I don't know why the time is indeterministic. But at least this pull request completely eliminates the PowerShell invocation and saves at least a couple of seconds in each checkout run.

Apart from the `windows-release` problem discovered by @dscho, another source of stall I discovered is extremely slow disk I/O for loading NodeJS and Git executable and libraries: https://github.com/actions/checkout/issues/1186#issuecomment-1484293074. But I don't think it's something can be improved by external users like us.

BrettDong:
> I could only reproduce this issue on larger runners, not on windows-latest or windows-2019. Do you experience it there too?

I never used large runners and I don't understand what issue you are referring to.

thboop:
Hey @BrettDong , thanks for putting this together, I'd love to get this merged and released.

It looks like the breaking change in the new major version of `actions/github` is causing this new behavior to fail. Do you have some time to update this pr, or should we take it over?

The underlying library octokit moved to put response fields into the `.data` subfield on the response. Since the hierarchy has changed, we need to make reaction changes in the areas where we call `actions/github` rest apis. You can see the original [patch notes](https://github.com/octokit/octokit.js/blob/f4845cffb392fcb0671d70e1e2779eabe0c8fd12/CHANGELOG.md#900)

BrettDong:
I did not run `npm build` last time and it might be the cause of test failures. @thboop: Could you approve CI checks?

src/github-api-helper.ts
```diff
@@ -125,19 +125,14 @@ async function downloadArchive(
   commit: string,
   baseUrl?: string
 ): Promise<Buffer> {
-  const octokit = getOctokit(authToken, {baseUrl: baseUrl})
-  const params: Octokit.ReposGetArchiveLinkParams = {
+  const octokit = github.getOctokit(authToken, {baseUrl: baseUrl})
+  const download = IS_WINDOWS
+    ? octokit.rest.repos.downloadZipballArchive
+    : octokit.rest.repos.downloadTarballArchive
+  const response = await download({
```
fhammerl, line 132: No need for handling the exception with something like `response.status != 200`. If the request fails, an exception is thrown and handled upstream.

image

fhammerl (approved):
LGTM!

fhammerl:
Thanks for your contribution @BrettDong 

Small side-note, did you need to manually recognise the MIT license of `.licenses/npm/@actions/http-client-1.0.8.dep.yml` or did `licensed` work the first time?

BrettDong:
> Thanks for your contribution @BrettDong
> 
> Small side-note, did you need to manually recognise the MIT license of `.licenses/npm/@actions/http-client-1.0.8.dep.yml` or did `licensed` work the first time?

I manually updated the yaml, following licensed documentation.

src/github-api-helper.ts
```diff
@@ -84,11 +84,11 @@ export async function getDefaultBranch(
 ): Promise<string> {
   return await retryHelper.execute(async () => {
     core.info('Retrieving the default branch name')
-    const octokit = getOctokit(authToken, {baseUrl: baseUrl})
+    const octokit = github.getOctokit(authToken, {baseUrl: baseUrl})
```
1newsr, line 87: The `baseUrl` must be converted to `serverApiUrl`.

I fixed this by https://github.com/actions/checkout/pull/1286.

BrettDong: Oh thanks!

ZauberNerd:
This PR breaks GHES compatibility in case `actions/checkout` is used in an environment where no `git` binary exists. See also: https://github.com/actions/checkout/pull/1288

PR #1248, opened 2023-03-26
Add to README how to checkout PR in `issue_comment` event

### LINKED ISSUE #331: Any way to checkout PR from `issue_comment` event?
The example in the docs [here](https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit) assumes a pull request event, but it does not work when operating on Pull Request comments, since those come through the `issue_comment` event.

I've spent several hours on this now and I don't see any way to checkout the branch associated with the issue_event.

Complications:

 - The event is the comment and it links to an issue, which only indirectly seems to have data on the pull request url.
 - Checking out the associate sha doesn't give push ability, since the branch name itself is not discoverable (at least not as well as I can tell).
 - Since `issue_comment` events have to be triggered by the workflow file on the default branch, it seems that is always the branch name provided. I haven't yet found any way to get the actual PR branch, although it seems this should be a straightforward and frequent use case for operating on PR comments.

ADTC:
Closes #331

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ When Git 2.18 or higher is not in your PATH, falls back to the REST API to downl
 - [Checkout multiple repos (private)](#Checkout-multiple-repos-private)
 - [Checkout pull request HEAD commit instead of merge commit](#Checkout-pull-request-HEAD-commit-instead-of-merge-commit)
 - [Checkout pull request on closed event](#Checkout-pull-request-on-closed-event)
+- [Checkout pull request on `issue_comment` event](#Checkout-pull-request-on-issue_comment-event)
 - [Push a commit using the built-in token](#Push-a-commit-using-the-built-in-token)
 
 ## Fetch all history for all tags and branches
@@ -210,6 +211,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+```
+
+## Checkout pull request on `issue_comment` event
+
+**Note:** The job must check if the issue is a pull request using `if: github.event.issue.pull_request` which skips it otherwise.
+
+```yaml
+- uses: actions/checkout@v3
+  with:
+    ref: refs/pull/${{ github.event.issue.number }}/head
 ```
 
 ## Push a commit using the built-in token


Spiritaine (approved):

Spiritaine (approved):

zyv:
I think the alternative with `gh` cli is worth mentioning, especially if afterwards one wants to do something with the branch.

ADTC:
@zyv is that the same as the latest comment in the issue? If not, please share your version here.

zyv:
@ADTC nope, I meant this option:

```yaml
- uses: actions/checkout@v3

- run: hub pr checkout ${{ github.event.issue.number }}
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Your option is nice, but it's only useful if you want to check out the code e.g. for static checks or something. The downside though is that if you check out the code your way, you can't use the resulting checkout to push the changes back to the branch with `git push` afterwards.

That's why I suggest to mention both alternatives if not the more generic one, and explain the caveat of your "nicer" solution, to spare folks trying to push their changes back banging the head against the wall for too long.

ADTC:
@zyv what's the functional difference between your version and the one [in this comment](https://github.com/actions/checkout/issues/331#issuecomment-1567674767)? Looks like they both do the same action, with yours being simpler code. Is that right?

Also, what's the difference between `gh pr` and `hub pr`?

zyv:
> @zyv what's the functional difference between your version and the one [in this comment](https://github.com/actions/checkout/issues/331#issuecomment-1567674767)? Looks like they both do the same action, with yours being simpler code. Is that right?

Correct.

> Also, what's the difference between `gh pr` and `hub pr`?

`gh` is better:

https://github.com/cli/cli/blob/trunk/docs/gh-vs-hub.md

I guess you can just replace `hub` with `gh` in my code, but I haven't tested that.

ADTC:
> I guess you can just replace `hub` with `gh` in my code, but I haven't tested that.

It will need to be tested. From what I gather, `gh` is **not** guaranteed to be backwards compatible with `hub`.

Also, are both `hub` and `gh` available by default in GitHub actions, or do you need to install them?

zyv:
> Also, are both `hub` and `gh` available by default in GitHub actions, or do you need to install them?

Yes, see:

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#preinstalled-software

Search for "GitHub CLI".

PR #1276, opened 2023-04-09
Set `user.signingKey` to `ssh-key`

### LINKED ISSUE #1239: Default `user.signingKey` to `ssh-key`
https://github.com/actions/checkout/blob/8f4b7f84864484a7bf31766abe9204da3cbe65b3/src/git-auth-helper.ts#L270

wojpawlik:
Slightly simplifies reusing ssh key for signing commits and tags. Doesn't change `gpg.format`, to avoid breaking users of gpg keys .

Closes #1239.


wojpawlik:
Leaving key management to a single action is more secure (and aesthetically pleasing :sweat_smile:)
image

PR #1288, opened 2023-04-13
Pass correct baseUrl to octokit

ZauberNerd:
The PR #1246 replaced the `getOctokit` method from the `octokit-provider.ts` file with the `getOctokit` method from the `@actions/github` package.
The octokit-provider was previously responsible for creating an Octokit instance and setting the `baseUrl` via the `getServerApiUrl` helper function. This function calls `getServerUrl` which reads the server url from the `GITHUB_SERVER_URL` environment variable, which on GHES is set to the enterprise instance.
This commit restores the previous behaviour by calling `getServerApiUrl` in all places where an octokit instance is created.

### DIFF
--- a/__test__/github-api-helper.test.ts
+++ b/__test__/github-api-helper.test.ts
@@ -1,0 +1,37 @@
+import * as github from '@actions/github'
+import * as githubApiHelper from '../lib/github-api-helper'
+
+jest.mock('@actions/github')
+
+describe('github-api-helper tests', () => {
+  describe('github enterprise compatibility', () => {
+    beforeEach(() => {
+      process.env.GITHUB_SERVER_URL = 'https://enterprise.git.com'
+    })
+
+    afterEach(() => {
+      delete process.env.GITHUB_SERVER_URL
+    })
+
+    it('getDefaultBranch should use GITHUB_SERVER_URL to set the baseUrl', async () => {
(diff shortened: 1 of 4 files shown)


__test__/github-api-helper.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    beforeEach(() => {
+      process.env.GITHUB_SERVER_URL = 'https://enterprise.git.com'
+    })
+
+    afterEach(() => {
+      delete process.env.GITHUB_SERVER_URL
+    })
+
+    it('getDefaultBranch should use GITHUB_SERVER_URL to set the baseUrl', async () => {
+      ;(github.getOctokit as jest.Mock).mockImplementation(() => {
```
fhammerl, line 17: Is the `;` necessary for this typecast?

ZauberNerd: Technically no, but the prettier configuration in this repository seems to require it. At least the `format-check` fails without it and `format` adds it.

fhammerl: Ah right, my mistake, it is correct. The `;` is to prevent the parentheses of `(github.getOctokit as jest.Mock)` to be interpreted as function invocation in case later statement is added above.

```
console.log('foo')
(github.getOctokit as jest.Mock)
```
would be read as `console.log('foo')(github.getOctokit as jest.Mock)`, the `;` prevents it.

Issue #800, opened 2022-05-21
Support Sparse checkouts


devminded:
### Discussed in https://github.com/actions/checkout/discussions/798

<div type='discussions-op-text'>

<sup>Originally posted by **devminded** May 20, 2022</sup>
# Support sparse checkouts

## Purpose
When using large project with many modules (AKA monorepo) being able to do sparse checkouts (supported since git 2.25.x) would be very helpful.

## Example
**Repo containing:**
- scripts
- tools
- modules/mobile-client
- modules/server
- modules/web

We want to only checkout and build the mobile-client (nice when combined with paths/paths_ignore).
So the equivalent of `git sparse-checkout init --cone && git sparse-checkout set modules/mobile-client scripts tools` would be:

```yaml
- name: Checkout tools repo
  uses: actions/checkout@v3
  with:
    repository: my-org/my-repo
    sparse:
      - modules/mobile-client
      - scripts
      - tools
```

Ref: https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/
</div>

devminded:
Related #172 #680

mversluys:
Worth noting that this https://github.com/actions/checkout/blob/main/src/git-source-provider.ts#L188 causes all LFS files to be retrieved for the given branch.

When using sparse-checkout *AND* lfs, checkout will only download lfs files as needed so the sparse rules can be used to prevent a lot of stuff being downloaded which is great, but only with the call to `git lfs fetch origin ...` being removed.

KIC:
any news on this? I this feature planned to be implemented?

dfdez:
Hey everyone! Good news for those that need this feature, I have created a [PR](https://github.com/actions/checkout/pull/1317) to add `sparse-checkout` support! 🎉

dfdez:
This Issue has been solved! https://github.com/actions/checkout/pull/1369

tamird:
This can probably be closed? @TingluoHuang

devminded:
Thank you very much team.

PR #1317, opened 2023-05-13
Add support for sparse checkout

dfdez:
# Context:
Monorepo sometimes can be challenging mainly when having a huge base of code. A lot of people have encounter situation where they don't need to fetch everything inside the repository, with this new option anyone will be able to make `sparse-checkout` inside github action quite simple!

# Features:
- Added `sparse-checkout` option
- Added test for `sparse-checkout`
- Updated `README.md` with `sparse-checkout` option and examples

# Usage:

## Fetch only the root files
```yaml
- uses: actions/checkout@v3
  with:
    sparse-checkout: .
```

## Fetch only the root files and `.github` and `src` folder

```yaml
- uses: actions/checkout@v3
  with:
    sparse-checkout: |
      .github
      src
```

> Once the first `sparse-checkout` is made you can download more folders with the [command](https://git-scm.com/docs/git-sparse-checkout#Documentation/git-sparse-checkout.txt-codegitsparse-checkoutaddSOMEDIRECTORYcode) `git sparse-checkout add SOME/DIR/ECTORY` more details rewarding `sparse-checkout` [here](https://git-scm.com/docs/git-sparse-checkout)

# Implementation

In order to make `sparase-checkout` much useful I have added to the `git fetch` the possibility to add a filter in order to use `--filter='blob:none'`.

With this update we will make sure that when the fetch is done no blob of the repository will be downloaded (this filter will be done only when `sparse-checkout` is set), instead the blobs will be downloaded when making the `git checkout` after having the `sparse-checkout` configured.

[Here](https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/) you can find the article that I have used to guide me to make the implementation.

# Related Issues
- https://github.com/actions/checkout/issues/800
- https://github.com/actions/checkout/pull/680

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,7 +117,7 @@ jobs:
       - name: Sparse checkout basic example
         uses: ./
         with:
-          sparse-checkout: .
+          sparse-checkout: __test__
       - name: Verify sparse checkout basic
         run: __test__/verify-sparse-checkout-basic.sh
 
@@ -126,6 +126,7 @@ jobs:
         uses: ./
         with:
           sparse-checkout: |
+            __test__
             .github
             src
       - name: Verify sparse checkout basic
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,20 @@ jobs:
         shell: bash
         run: __test__/verify-side-by-side.sh
 
+     # Sparse checkout
+      - name: Sparse checkout
+        uses: ./
+        with:
+          sparse-checkout: |
+            __test__
+            .github
+            dist
+
+      - name: Verify sparse checkout basic
+        run: __test__/verify-sparse-checkout-basic.sh
+      - name: Verify sparse checkout example
+        run: __test__/verify-sparse-checkout.sh
+
       # LFS
       - name: Checkout LFS
         uses: ./
@@ -112,19 +126,6 @@ jobs:
           submodules: recursive
       - name: Verify submodules recursive
         run: __test__/verify-submodules-recursive.sh
-
-     # Sparse checkout
-      - name: Sparse checkout
-        uses: ./
-        with:
-          sparse-checkout: |
-            __test__
-            .github
-            dist
-      - name: Verify sparse checkout basic
-        run: __test__/verify-sparse-checkout-basic.sh
-      - name: Verify sparse checkout example
-        run: __test__/verify-sparse-checkout.sh
 
       # Basic checkout using REST API
       - name: Remove basic
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,6 +113,26 @@ jobs:
       - name: Verify submodules recursive
         run: __test__/verify-submodules-recursive.sh
 
+      # Sparse checkout basic example
+      - name: Sparse checkout basic example
+        uses: ./
+        with:
+          sparse-checkout: .
+      - name: Verify sparse checkout basic
+        run: __test__/verify-sparse-checkout-basic.sh
+
+     # Sparse checkout example
+      - name: Sparse checkout example
+        uses: ./
+        with:
+          sparse-checkout: |
+            .github
+            src
+      - name: Verify sparse checkout basic
+        run: __test__/verify-sparse-checkout-basic.sh
+      - name: Verify sparse checkout example
+        run: __test__/verify-sparse-checkout.sh
+
       # Basic checkout using REST API
       - name: Remove basic
         if: runner.os != 'windows'
@@ -205,7 +225,7 @@ jobs:
           path: basic
       - name: Verify basic
         run: __test__/verify-basic.sh --archive
-    
+
   test-git-container:
     runs-on: ubuntu-latest
     container: bitnami/git:latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,22 +113,14 @@ jobs:
       - name: Verify submodules recursive
         run: __test__/verify-submodules-recursive.sh
 
-      # Sparse checkout basic example
-      - name: Sparse checkout basic example
-        uses: ./
-        with:
-          sparse-checkout: __test__
-      - name: Verify sparse checkout basic
-        run: __test__/verify-sparse-checkout-basic.sh
-
-     # Sparse checkout example
-      - name: Sparse checkout example
+     # Sparse checkout
+      - name: Sparse checkout
         uses: ./
         with:
           sparse-checkout: |
             __test__
             .github
-            src
+            dist
       - name: Verify sparse checkout basic
         run: __test__/verify-sparse-checkout-basic.sh
       - name: Verify sparse checkout example
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,6 +80,7 @@ jobs:
             __test__
             .github
             dist
+          path: sparse-checkout
 
       - name: Verify sparse checkout basic
         run: __test__/verify-sparse-checkout-basic.sh
--- a/README.md
+++ b/README.md
@@ -74,6 +74,10 @@ When Git 2.18 or higher is not in your PATH, falls back to the REST API to downl
     # Default: true
     clean: ''
 
+    # Do a sparse checkout on given patterns (each pattern should be sepparated with new lines).
+    # Default: null
+    sparse-checkout: ''
+
     # Number of commits to fetch. 0 indicates all history for all branches and tags.
     # Default: 1
     fetch-depth: ''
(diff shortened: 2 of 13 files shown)


blu3mania:
Same comment as I posted on #680:

Shouldn't lfs.fetchinclude also be set or use "-I" (or "--include") parameter in GitCommandManager.lfsFetch to make sure LFS only fetches from included paths?

Spiritaine (approved):

dfdez:
> Shouldn't lfs.fetchinclude also be set or use "-I" (or "--include") parameter in GitCommandManager.lfsFetch to make sure LFS only fetches from included paths?

Regarding this topic and after taking a look and  making some test using "--include" option directly when making the `git lfs fetch` will end up in a first try error:

 image
 
 > I think this happens because with `sparse-checkout` we have make the `fetch` with `--filter=blob:none` therefore `lfs` is missing information

In order to have working `lfs` correctly with `sparse-checkout` I thought in 2 possible solutions:
1. Skip the `lfs fetch` step and don't set `GIT_LFS_SKIP_SMUDGE=1` and fetch will be done in the actual `checkout`
2. Skip the `lfs fetch` step and run a `git lfs pull` after `checkout` with "--include" option

Let me know what you think about this! But my personal opinion is that the second solution is the  best one.

dscho:
> Skip the `lfs fetch` step and run a `git lfs pull` after `checkout` with "--include" option

Disclaimer: I am not using Git LFS myself. Having said that, I _guess_ that would allow Git LFS to batch-fetch, and the first option maybe would not? If so, I am definitely in favor of the second option. Even if both allow batch-fetching, the second option looks slightly cleaner to me. Probably something like this?

```diff
diff --git a/__test__/git-auth-helper.test.ts b/__test__/git-auth-helper.test.ts
index fec6573..fa2d1db 100644
--- a/__test__/git-auth-helper.test.ts
+++ b/__test__/git-auth-helper.test.ts
@@ -759,6 +759,7 @@ async function setup(testName: string): Promise<void> {
     init: jest.fn(),
     isDetached: jest.fn(),
     lfsFetch: jest.fn(),
+    lfsPull: jest.fn(),
     lfsInstall: jest.fn(),
     log1: jest.fn(),
     remoteAdd: jest.fn(),
diff --git a/__test__/git-directory-helper.test.ts b/__test__/git-directory-helper.test.ts
index 362133f..9d9d8bd 100644
--- a/__test__/git-directory-helper.test.ts
+++ b/__test__/git-directory-helper.test.ts
@@ -474,6 +474,7 @@ async function setup(testName: string): Promise<void> {
     init: jest.fn(),
     isDetached: jest.fn(),
     lfsFetch: jest.fn(),
+    lfsPull: jest.fn(),
     lfsInstall: jest.fn(),
     log1: jest.fn(),
     remoteAdd: jest.fn(),
diff --git a/src/git-command-manager.ts b/src/git-command-manager.ts
index 4f6dc79..89b3a44 100644
--- a/src/git-command-manager.ts
+++ b/src/git-command-manager.ts
@@ -40,6 +40,7 @@ export interface IGitCommandManager {
   init(): Promise<void>
   isDetached(): Promise<boolean>
   lfsFetch(ref: string): Promise<void>
+  lfsPull(): Promise<void>
   lfsInstall(): Promise<void>
   log1(format?: string): Promise<string>
   remoteAdd(remoteName: string, remoteUrl: string): Promise<void>
@@ -328,6 +329,15 @@ class GitCommandManager {
     })
   }
 
+  async lfsPull(): Promise<void> {
+    const args = ['lfs', 'pull', '--include']
+
+    const that = this
+    await retryHelper.execute(async () => {
+      await that.execGit(args)
+    })
+  }
+
   async lfsInstall(): Promise<void> {
     await this.execGit(['lfs', 'install', '--local'])
   }
diff --git a/src/git-source-provider.ts b/src/git-source-provider.ts
index 967097d..e8e3e44 100644
--- a/src/git-source-provider.ts
+++ b/src/git-source-provider.ts
@@ -188,7 +188,8 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
     // LFS fetch
     // Explicit lfs-fetch to avoid slow checkout (fetches one lfs object at a time).
     // Explicit lfs fetch will fetch lfs objects in parallel.
-    if (settings.lfs) {
+    // For sparse checkouts, wait until after the checkout is done.
+    if (settings.lfs && !settings.sparseCheckout) {
       core.startGroup('Fetching LFS objects')
       await git.lfsFetch(checkoutInfo.startPoint || checkoutInfo.ref)
       core.endGroup()
@@ -210,6 +211,13 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
     await git.checkout(checkoutInfo.ref, checkoutInfo.startPoint)
     core.endGroup()
 
+    // Sparse checkout only: delayed LFS pull
+    if (settings.lfs && settings.sparseCheckout) {
+      core.startGroup('Pulling LFS objects')
+      await git.lfsPull()
+      core.endGroup()
+    }
+
     // Submodules
     if (settings.submodules) {
       // Temporarily override global config
```

dscho:
Hmm. I cannot get that `git lfs pull --include` thing to work, essentially because of the arbitrary patterns allowed in sparse-checkout... So I am going to try option 1.

dfdez:
@dscho the "--include" pattern should work the same as the `sparse-checkout` syntax that shouldn't be a problem If you want I will try option 2 and you try option 1?

blu3mania:
@dfdez @dscho Are you guys sure "git lfs -I" does not support patterns? This document seems to suggest it does: https://manpages.debian.org/testing/git-lfs/git-lfs-fetch.1.en.html. I also searched git-lfs repo and found this issue: git-lfs/git-lfs#912, which clearly suggest that patterns should be supported. If possible, we should use lfs pull for better performance. I don't use sparse checkout in dev environment, but our Jenkins build does, so I took a look at the build job log, and could confirm that it used lfs.fetchinlcude config and "lfs pull" to do so:

```
12:04:26 Cloning repository git@github.com:org_redacted/repo_redacted.git
12:04:26  > git.exe init C:\Jenkins\workspace\JobName # timeout=10
12:04:27 Fetching upstream changes from git@github.com:org_redacted/repo_redacted.git
12:04:27  > git.exe --version # timeout=10
12:04:27  > git --version # 'git version 2.41.0.windows.1'
12:04:27 using GIT_SSH to set credentials GitHub account SSH key
12:04:27 Verifying host key using known hosts file
12:04:27  > git.exe fetch --no-tags --force --progress -- git@github.com:org_redacted/repo_redacted.git +refs/heads/*:refs/remotes/origin/* # timeout=120
12:55:49  > git.exe config remote.origin.url git@github.com:org_redacted/repo_redacted.git # timeout=10
12:55:49  > git.exe config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/* # timeout=10
12:55:49 Avoid second fetch
12:55:49  > git.exe rev-parse "origin/branch_redacted^{commit}" # timeout=10
12:55:49 Checking out Revision rev_redacted (origin/branch_redacted)
12:55:49 Enabling Git LFS pull
12:55:49  > git.exe config core.sparsecheckout # timeout=10
12:55:49  > git.exe config core.sparsecheckout true # timeout=10
12:55:49  > git.exe config lfs.fetchinclude /folder_redacted # timeout=10
12:55:49  > git.exe config --unset lfs.fetchexclude # timeout=10
12:55:49  > git.exe read-tree -mu HEAD # timeout=10
12:55:49  > git.exe checkout -f rev_redacted # timeout=120
12:56:31  > git.exe config --get remote.origin.url # timeout=10
12:56:31 using GIT_SSH to set credentials GitHub account SSH key
12:56:31 Verifying host key using known hosts file
12:56:31  > git.exe lfs pull origin # timeout=120
```

Note that somehow the log didn't show how it performed sparse checkout, but I confirmed that it used .git/info/sparse-checkout file to specify the sparse checkout path. Though, it doesn't matter since "git sparse-checkout" should be doing the same.

dscho:
> Are you guys sure "git lfs -I" does not support patterns?

The manual page talks about `<paths>`, not `<patterns>`. Which is misleading, granted, given that it says that wildcard matching as per gitignore is applied. However, there are _still_ two rather big concerns using `git lfs pull -i`:

- the `sparse-checkout` file contains directory names (in cone mode), which would potentially need to be translated into proper wildcard patterns by appending `/*`.
- while the documentation refers to "wildcard matching as per `.gitignore`", the `--include` option uses a _comma-separated list_ of paths, meaning that the comma is actually handled differently from `.gitignore` (or for that matter, `sparse-checkout`). Therefore, if the `sparse-checkout` definition contains, say, `/blu3,mania/`, a simple implementation using that option would not work correctly.
- these two issues are not exactly big blockers, but demonstrate that it would be far from trivial to use the `-i` option.
- long sparse-checkout definitions could run into command-line limitations.

With all of that, I actually really do not want to use this option.

> we should use "lfs pull" for better performance

Are you sure about that? From what I gather reading https://git-scm.com/docs/gitattributes#_delay, I could imagine that it works as intended (i.e. using batched LFS fetching) even when using sparse checkouts.

As I mentioned earlier, I am by no means familiar with LFS. I don't have access to any meaningful repository using LFS. Maybe you do, and maybe you can test whether letting `git checkout` fetch the LFS objects in sparse checkouts lets them be fetched individually, or in batches?

blu3mania:
Seems you are right. I don't have the env at home so I didn't test but after some more digging, it appears that "git lfs pull has better performance" was just an old myth. According to https://github.com/git-lfs/git-lfs/issues/2594#issuecomment-330363293 it shouldn't be a problem for Git anymore after Git 2.15. So your 25d6c12 should be the right solution. Thanks.

dfdez:
> we should use "lfs pull" for better performance

After some digging I found out [this](https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-clone.adoc#description) on the git-lfs docs, but as @blu3mania menthioned https://github.com/git-lfs/git-lfs/issues/2594#issuecomment-330363293 clarify everything!

Therefore we can close the topic regarding `lfs` and `sparse-checkout`! Thanks both!

dfdez:
Merged [here](https://github.com/actions/checkout/pull/1369) thanks @dscho for the help!

dscho:
> Merged [here](https://github.com/actions/checkout/pull/1369) thanks @dscho for the help!

@dfdez this was for the most part your work, culminating in https://github.com/actions/checkout/releases/tag/v3.5.3. Thank you so much! If I was able to assist, I am glad that I could.

dfdez:
> @dfdez this was for the most part your work, culminating in https://github.com/actions/checkout/releases/tag/v3.5.3.

Yeah that's true and I would have appreciated a small mention in the release 😅 

But the important thing is to have the functionality! 

Thanks for everything!

dscho:
@dfdez you're absolutely correct, your contribution must be celebrated. Look at the latest revision of the release notes of https://github.com/actions/checkout/releases/tag/v3.5.3 😃

dfdez:
Thanks! 🚀

Issue #270, opened 2020-06-02
Why do I have to pull the lfs separately when using lfs: 'true'


gary-cowell:
```
                  - name: checkout
                    uses: actions/checkout@v2
                    with:
                        lfs: 'true'
                  - name: checkoutLFS
                    uses: actions/checkout@v2
                  - run: git lfs pull
```

With only the first of those, the source checkout has LFS which look like this:

`-rw-r--r-- 1 runner docker 133 Jun  1 15:54 largefile.zip`

I.e. that's a reference only.

After the extra pull, I see:

 `-rw-r--r-- 1 runner docker 28956993 Jun  2 07:08 largefile.zip`

The lfs: 'true' is having some effect on the first checkout because I can see it installing LFS and issuing:

`/usr/bin/git lfs fetch origin refs/remotes/origin/mybranch`

That does not appear to take enough time to be fetching the actual lfs, then it runs

`/usr/bin/git checkout --progress --force -B mybranch refs/remotes/origin/mybranch`

Which still leaves us with a reference file.

It's only when I run the second checkout and force a 'git lfs pull' do I get the actual file. I don't see any errors in the first checkout step.

mcraiha:
I am also wondering this.

maximveksler:
Any input on this? Is this the expected behavior?

yair-mantis:
mybe this is no longer the case, at least on ubuntu[1]. 
using ssh key as secret. 
i use an 'ls'  on one of the files after the checkout to verify

name: CI 

on:
  workflow_dispatch:
```yaml
jobs:
  build:
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@v2.3.4
        with:
          lfs: 'true'
          submodules: 'recursive'
          ssh-key: ${{ secrets.git_ssh_key }}

        env:
          GIT_CURL_VERBOSE: '1'
          GIT_TRACE: '1'
```

[1] #311

bk2204:
Just popping in here since someone mentioned this on the Git LFS issue tracker, but are you sure the file you're seeing problems with is properly listed in `.gitattributes`, including case?  For example, Git will not think that `*.ZIP` in `.gitattributes` matches `largefile.zip` or vice versa.  `git lfs pull` may have nonstandard behavior here.

Also, do you have a public repo that demonstrates this problem?  That would be helpful to help whomever is looking at this troubleshoot the problem.

nevyn:
https://github.com/alloverse/alloplace

[...]

PR #1392, opened 2023-06-27
fix(lfs): ensure that LFS objects are checked out

sxlijin:
The current implementation of `lfs: true` does not actually guarantee that LFS objects are checked out:

* it first does a [`git lfs fetch origin`](https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-fetch.adoc) (which does _not_ update the working copy),
* and then relies on `git checkout` in combination with the smudge filter to check LFS objects out.

However, there are numerous reports of flakiness with this approach (and we've seen this ourselves in our own usage); see #270.

Adding a [`git lfs checkout` ](https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-checkout.adoc) should make this more robust.

Unfortunately I don't have a consistent repro for this issue, beyond the anecdata from our own workflow failures.


ceisen1:
Any particular reason why you closed this PR @sxlijin?

sxlijin:
Hmm... I thought we had internally come to a different conclusion about this, but double checking our CI setup we didn't.

One of the folks on my team had the hypothesis that all the flakiness issues people have reported have to do with persistent/shared state, since we use persistent self-hosted runners.

ceisen1:
Got it. That makes sense based on your description. But I do believe your original comments are correct that only running fetch is never going to update the lfs files in the working copy.

My one suggestion for your PR would be to move the git lfs checkout call into an if block where `settings.lfs`  is True

Ravio1i:
Hey there, we do face the same issue, on windows we have to run

```yaml
   - uses: actions/checkout@v3
      with:
        lfs: true

    - name: Git LFS Pull 
      run: git lfs pull
```

to ensure it actually gets pulled

genya0407:
Hi, I am facing the same issue. Are there any plans to merge this PR?

Issue #1152, opened 2023-02-01
Support for Treeless clones


tiger12506:
There is already fetch-depth: 1 to retrieve only the latest commit and working tree, which is great.
However, for my particular CI project using Actions, we use the git tags to track version information, and the commit messages to generate changelogs. Seems like the feature "treeless clones" would be ideal for our situation.

I couldn't figure out how to make this work with actions/checkout@v3, so I assume that support for it would need to be added.

https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/
![image](https://user-images.githubusercontent.com/1036922/216177628-43f751e9-8dda-41bd-96ab-a180ab18f541.png)

j-liedtke:
This feature would be very helpfull indeed

prein:
Related https://github.com/actions/checkout/issues/172 and https://github.com/actions/checkout/issues/663 and

marc-hb:
Thanks for filing this! 

> There is already fetch-depth: 1 to retrieve only the latest commit and working tree, which is great.

Actually, shallow clones are "great" until you start trying to use git; then they can turn into disasters. This is the worst git disaster I found:
- https://github.com/thesofproject/linux/issues/2556

Even `git describe` is changed which means you can't https://reproducible-builds.org/

So the (newer?) `--filter=tree:0` option seems  much, much better from a CI perspective than shallow clones because it doesn't seem to "break git" so far and the cloning speed-up seems pretty dramatic, 2-3 times faster in my experience. Shallow cloning is not faster, maybe even slower, see some numbers here:
- https://github.com/zephyrproject-rtos/west/issues/319

marc-hb:
@derrickstolee your (excellent!) blog and work about this is now 1.5 years old. It's a largely wasted effort if the universal way to clone on github still can't use these optimizations.

derrickstolee:
One thing to be really careful about is the fact that fetches from treeless clones can be very strange if there is a `.gitmodules` file in the repo. That will really only affect users that run `git fetch` within their workflow for some reason. I mentioned this in the article:

[...]

Issue #663, opened 2021-12-28
Checkout small files only


pcgeek86:
I have some large files in my repository that are slowing down the checkout process. To work around this locally, I can use the following command:

`git clone --no-checkout --filter=blob:limit=1m`

However, I cannot find any ways to pass in these additional options to the `checkout` action.

Any ideas how to skip large files with this checkout action?

jakub-g:
Note GitLab supports `GIT_FETCH_EXTRA_FLAGS` which can be used for that purpose:

- https://docs.gitlab.com/ee/ci/large_repositories/
- https://docs.gitlab.com/ee/ci/runners/configure_runners.html#git-fetch-extra-flags

flying-sheep:
As a hack, I assume one could do the following:

```yaml
# sparse checkout to set the blob:none filter, but include all files
sparse-checkout: '/*/'
# allow interacting with the Git history
fetch-depth: 0
```

That way, we get a clone with intact history but no old blobs being downloaded.

A real fix is in https://github.com/actions/checkout/pull/1396

PR #1396, opened 2023-06-30
Add support for partial checkout filters

finleygn:
A change to add support for sparse checkouts has been added recently (#1369). This extends upon that to allow support for other filters, and the ability to filter without a sparse checkout.

```yaml
- uses: actions/checkout@v3
  with:
    filter: tree:0
```

Inspired by this blog post -> https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone


Spiritaine (approved):

gabrielyotoo:
bump +1 this would be nice to use with Yarn

petergaultney:
i've used this fork/branch successfully with both `tree:0` and `blob:none`. Thank you!

flying-sheep:
I think `filter: 'tree:0'` would be a much better default than `fetch-depth: 1`.

It would be almost exactly as small, but without breaking the Git history.

__test__/verify-fetch-filter.sh
```diff
@@ -0,0 +1,15 @@
...
+  echo "Expected ./fetch-filter/.git folder to exist"
+  exit 1
+fi
+
+# Verify .git/config contains partialclonefilter
+
+CLONE_FILTER=$(git config --local --get remote.origin.partialclonefilter)
+
+if [ "$CLONE_FILTER" != "blob:none" ]; then
+  echo "Expected ./fetch-filter/.git/config to have 'remote.origin.partialclonefilter' set to 'blob:none'"
```
gabrielyotoo, line 14: did you miss exit 1 here?

finleygn: yep, good spot 😳

kalanzai:
This is exactly what we need in our repo. Speeds up checkout by a factor of 4 without breaking git history so our existing logic that uses `git diff` and `git blame` will still work ❤️

cory-miller:
Thanks for your contribution.

Do you mind adding an update to https://github.com/actions/checkout/blob/72f2cec99f417b1a1c5e2e88945068983b7965f9/action.yml#L56C19-L60 with the new setting with a note that it will override the `sparseCheckout` option if also set?

The rest LGTM.

Spiritaine (approved):

finleygn:
Will take a look as to whats causing the test to fail soon...

finleygn:
@cory-miller the test actions *should* pass now 🙏

cory-miller (approved):

cory-miller:
Thanks for working on this!

I will build a release soon.

cory-miller:
`actions/checkout@v4` or `actions/checkout@v4.1.0` are updated with this change.

Rycieos:
@finleygn thanks for working on this!

@cory-miller the issues #172, #663, and #1152 should likely be closed.

benoit-pierre:
This does not seem to apply to submodules.

Kurt-von-Laven:
> I think `filter: 'tree:0'` would be a much better default than `fetch-depth: 1`.
> 
> It would be almost exactly as small, but without breaking the Git history.

This is a cool feature, but please don't make the default checkout larger for everyone since most users don't need the Git history in their CI pipelines. We do often need the history, so we tried switching from `fetch-depth: 0` to `filter: tree:0`, and our CI pipelines became twice as slow _overall_. Then we tried `filter: blob:none`, and that still slowed down our CI pipelines by 50%. I suspect this is happening because some tool (e.g., Gitleaks) is causing Git to fetch the missing data piecemeal rather than more efficiently all at once by attempting to access trees or blobs that aren't available locally. I don't mean this as a knock on partial clones; everyone should always measure the impact of performance changes in their own specific use case. I would guess that treeless clones are a net performance benefit if and only if no trees or blobs from previous commits are needed, because GitHub has already put so much effort into optimizing checkout times in GitHub Actions. Similarly, I would guess that blobless clones are a net performance benefit if and only if no blobs from previous commits are needed.

jakebailey:
I think there's something missing from the docs here; you still need to explicitly set `fetch-depth: 0` in order to get all of history, making a full workflow look like:

```yml
- use: actions/checkout@v4
  with:
    fetch-depth: 0
    filter: blob:none
```

Without setting it, you still only get the most recent commit.

For the TypeScript repo:

- `fetch-depth: null`, `filter: null`: 12s, but only one commit.
- `fetch-depth: null`, `filter: blob:none`: 30s, but still only one commit!
- `fetch-depth: 0`, `filter: null`, 4m6s: full history
- `fetch-depth: 0`, `filter: filter:none`, 1m37s: full history

Kurt-von-Laven:
@jakebailey, did you mean `filter: blob:none` rather than `filter: filter:none` for the last bullet?

jakebailey:
Yes, that's just a typo on my part.

Issue #687, opened 2022-02-04
50Gb repo


blackliner:
My org is currently using Jenkins, and our monorepo is huge (above 50Gb on HEAD). On Jenkins, we utilize [git reference feature](https://argus-sec.com/how-you-can-use-git-reference-repository-to-reduce-jenkins-disk-space/) to limit bandwith usage. We are now evaluating different SaaS CI providers and this seems to be one of the big show-stoppers for most. So:

- What is the canonical approach on github actions for such extreme cases?
- Do (custom) runners support local references? (Enterprise use-case)
- Alternatives?

## References
https://github.com/actions/checkout/issues/22 is one issue I found, but I think that was just about fetch depth. Does not help, because even with a depth of 1, we would have to clone 50Gb.

blackliner:
something like https://circleci.com/docs/2.0/caching/#source-caching maybe?

Issue #22, opened 2019-08-24
suggestions for large repository


softprops:
I'm experimenting with Github Actions on a large code repository at `$WORK`. We use a mix of Concourse ci and Jenkins and are looking towards Github Actions as a potential CI/CD tool with less hosting maintenance. 

In some simple experiments we found the actions/checkout command to be slower that expected. A checkout step takes `~6m 53s` using `with.fetch-depth: 1`, and that's before my job can do anything useful.

In our Jenkins set up we have a persistent clone that we do local clones from. In concourse we use the [git resource](https://github.com/concourse/git-resource). In both cases the process of fetching a given version of the code feels faster than actions/checkout and was wondering if there as any actions tuning parameters would could apply to speed up the process

I can see a glimpse into what the action is doing in the output.

```
git remote add origin https://github.com/{org}/{repo}
 git config gc.auto 0
git config --get-all http.https://github.com/{org}/{repo}.extraheader
git config --get-all http.proxy
git -c http.extraheader="AUTHORIZATION: basic ***" fetch --tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/*:refs/remotes/origin/*
```
Is there a way we can opt out of features like fetching all of the tags (we have a lot of tags) or submodules (we don't use submodules)?

I started looking at want concourse git resource does to draw some comparisons. Looking [here](https://github.com/concourse/git-resource/blob/d99b8e5e824f0c4ee4aa096b2ce83894bcf0d2c3/assets/in#L81-L91) I can see it does something like a clone +  checkout operation

```sh
git clone --single-branch $depthflag $uri $branchflag $destination $tagflag

cd $destination

git fetch origin refs/notes/*:refs/notes/* $tagflag

if [ "$depth" -gt 0 ]; then
  "$bin_dir"/deepen_shallow_clone_until_ref_is_found_then_check_out "$depth" "$ref" "$tagflag"
else
  git checkout -q "$ref"
fi
```

I'm not sure if that's any better or worse our experience is that it "feels" more performance that what we see with github actions/checkout.

zoispag:
I have similar problem.
Is there a way to do something like `git clone --single-branch --branch ${{ github.head_ref }} --depth 1 https://github.com/${{ github.repository }}.git` ?

softprops:
@zoispag just curious what was the scope applied to `secrets. CLONE_TOKEN` vs the default GITHUB_TOKEN that GitHub provides to each action?

zoispag:
[...]

PR #1400, opened 2023-07-03
Add support for reference repository parameter

ddompe:
Git clone command added support for a local reference repository in order to minimize the network traffic and load on the git servers for big repositories. Adding support to specify such reference git repository in the checkout action.

The is very useful in big repositories hosted in GitHub Enterprise deployments that runs massive scales of CI jobs across distributed geographies.

Tested manually in my private GitHub Enterprise instances, but I lack the knowledge to create a unit test for this. Sharing to upstream so it can be taken from here, or pulled by other people needing it (see #22 and #687)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,11 @@ When Git 2.18 or higher is not in your PATH, falls back to the REST API to downl
     # Default: 1
     fetch-depth: ''
 
+    # Path to a local bare git [reference repository to minimize network
+    # usage](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---reference-if-ableltrepositorygt).
+    # If the directory doesn't exists this option will ignore it and log a message.
+    reference: ''
+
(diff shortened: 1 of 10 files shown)


bitcoinfinancier (approved):

Spiritaine (approved):

Pospelove:
does this work for git lfs?

tbranch227 (approved):
This looks great.  This feature really needs to be supported by this action.

tbranch227:
> does this work for git lfs?

I believe it does. We use --reference with our own processes (not this action) and it's a big performance improvement and offloads work from the server and WAN links.

sizehnde:
Bumping this PR. It would greatly benefit certain monorepo workflows.

We have implemented an Action internally to support reference clones that essentially has to reimplement the portions of `actions/checkout` we care about; it would be nice for this to just be a built in feature.

PR #1436, opened 2023-08-23
Update default runtime to node20

takost:
Node12 was deleted from runner. Node20 was added to Actions Runner on v2.308.0.
Node16 has en end of life on 11 Sep 2023.

This PR updates the default runtime to node20, rather then node16


Spiritaine (approved):

Link- (approved):
🚢

04131988 (approved):

zkochan:
It looks like this change was considered a breaking change, right? May I know why? Isn't this only changing the runtime of the `checkout` action? Hence, the user's code will still run on the node.js runtime configure by the user.

action.yml
```diff
@@ -87,6 +87,6 @@ inputs:
     description: The base URL for the GitHub instance that you are trying to clone from, will use environment defaults to fetch from the same instance that the workflow is running from unless specified. Example URLs are https://github.com or https://my-ghes-server.example.com
     required: false
```
HUFF-ANDERSON, line 88: Exe.run

920506837 (commented):
takost/update-node20

kikeavril16:
Alguien me explica que es esta app y que son los códigos y las solicitudes de extracción?? Porque estoy aquí y ni se para que sirve esto

920506837:
1436

Issue #1453, opened 2023-09-04
v4 with default config (no `with`) does not include --progress flag


bquorning:
After upgrading to actions/checkout@v4 (no `with` configuration), we see[^1] `show-progress: true` listed in the logs, but `git fetch` is run without the `--progress` flag.

<img width="869" alt="image" src="https://github.com/actions/checkout/assets/22333/354c438a-1012-4e87-a045-554a7d284b7b">

I think this is not the intended behavior of #1067, which says

> Note that the default value of this new option is `true`, which matches the current behavior, so this PR introduces no behavior change unless the new option is explicitly present and set to false.

[^1]: https://github.com/rubocop/rubocop-rspec/actions/runs/6076842370/job/16485526462

riseshia:
#1454 might fix this

nedbat:
I understand the desire to keep old behavior, but why does anyone need to see dozens of lines of progress percentages?  Since @v4 seems to have changed the default (by accident), why no just admit that no progress is the better default and leave it that way?

bquorning:
I too prefer the less noisy (accidental) default.

simonbaird:
Ooh, sorry about that. I think the original version of #1067 predates the new options handling and somehow when rebasing and resolving conflicts I didn't notice the way the default wasn't preserved. (In hindsight an inverted option might have been better, e.g. "hideProgress" instead of "showProgress".)

cbandera:
Hi what is the status of this issue? Has it been fixed in the meanwhile? I'm not sure whether I understand your comments but the referneced pr #1454 does not seem to have been merged.

riseshia:
Based on the threads so far, I concluded that maintainers is not interested in this bug, and I don't want to keep a PR opened long long time, so closed #1454 a few days ago. I still have willing to contribute for this if requested, but leave this for more passionate someone.

tianon:
Maybe resurrect #1454 but swap the input default to `false` instead of `true`?  or drop the (unused/unpassed) input instead?

(I got here by reading the `actions/checkout` code and being flabbergasted that the `showProgress` variable is preserved and carefully passed down through the layers and that there's even a place to set it on the `git.fetch` options argument, but that it simply ... doesn't get passed down. 😅)

PR #1454, opened 2023-09-04
Pass showProgress to fetchOptions

riseshia:
Fix: #1453

#1067 seems not to pass `showProgress` to fetchOptions, which makes any fetch call handled without `--progress`
Confirmed works correctly with default / explicitly given show-progress option, here is test result
https://github.com/riseshia/actions-playground/actions/runs/6078541181

I don't add test for this, because I couldn't find proper place for this case(most test are unit testing, but this problem is from "passing" arg) 🤔 
any suggestion for adding test is welcome. ;)

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1243,7 +1243,9 @@ function getSource(settings) {
             }
             // Fetch
             core.startGroup('Fetching the repository');
-            const fetchOptions = {};
+            const fetchOptions = {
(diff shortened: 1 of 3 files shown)


takost (approved):

riseshia:
Conflict resolved. Could anybody check this?
📝  diff for https://github.com/actions/checkout/issues/1455 is not included, we might handle this seperately.

PR #1482, opened 2023-09-18
Add missing id of diff step for check-dist workflow

### LINKED ISSUE #1480: missing step id in check-dist workflow
```yaml
      - name: Compare the expected and actual dist/ directories
        run: |
          if [ "$(git diff --ignore-space-at-eol dist/ | wc -l)" -gt "0" ]; then
            echo "Detected uncommitted changes after build.  See status below:"
            git diff
            exit 1
          fi

      # If dist/ was different than expected, upload the expected version as an artifact
      - uses: actions/upload-artifact@v2
        if: ${{ failure() && steps.diff.conclusion == 'failure' }}
        with:
          name: dist
          path: dist/
```

the last step uses `steps.diff.conclusion`, but there is no id defined.

islishude:
close #1480

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -36,6 +36,7 @@ jobs:
         run: npm run build
 
       - name: Compare the expected and actual dist/ directories
+        id: diff
         run: |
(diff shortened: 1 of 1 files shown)


islishude:
it's a waste of my time.

I won't contribute commits to github again.

PR #1493, opened 2023-09-21
Add support to customize working directory

accodev:
Allow the users to customize the working directory where the commands of the action will be executed, e.g.: c:\, /home/user, etc.
This is to allow the users to switch to a different mount point (for example github hosted windows runners offer quite a bit of space but the checkout process happens on d:\, where the space is only 14GB).
Sometimes it is not feasible to copy the repository after cloning it because:
- It might be too big, costing a lot of time and thus increasing the billing
- It might have constraints (symbolic links) that make it impossible to copy
I have added a few unit tests and tested it on a windows-latest runner on this repository: 
https://github.com/accodev/trackmania-manager-py/actions/runs/6263454133/job/17007891987
Note that the chosen working directory has to exists (kept the original behaviour).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Please refer to the [release page](https://github.com/actions/checkout/releases/
     # Default: true
     persist-credentials: ''
 
-    # Relative path under $GITHUB_WORKSPACE to place the repository
+    # Relative path under $GITHUB_WORKSPACE/working-directory to place the repository
     path: ''
 
     # Whether to execute `git clean -ffdx && git reset --hard HEAD` before fetching
@@ -126,6 +126,10 @@ Please refer to the [release page](https://github.com/actions/checkout/releases/
     # running from unless specified. Example URLs are https://github.com or
     # https://my-ghes-server.example.com
     github-server-url: ''
+
+    # Provide the working directory for the git commands to execute into, defaults to
+    # $GITHUB_WORKSPACE
+    working-directory: ''
 ```
 <!-- end usage -->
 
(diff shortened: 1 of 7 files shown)


dasigvr:
Yes! This is badly needed in our use case.

accodev:
Any chance this could get a review?

johnandersen777 (approved):
LGTM

snps-midhunn:
Hello,
Is there any update on when we can expect this to be part of the official release? We are very much in need of this feature too.

interifter (approved):

Rsnipeu:
Merge

Rsnipeu (approved):
Done

Rsnipeu (approved):
Merge

Pospelove (approved):
Go gogo gog

WiehanVanNiekerk:
Is there any tracking regarding this PR?

WiehanVanNiekerk (approved):
It looks all good for me. Do you know if we can get this in? We need this for our enterprise org repo's workflows.

WiehanVanNiekerk:
@cory-miller @takost @luketomlinson

WiehanVanNiekerk:
@jww3 Please can you assist

interifter:
Related/possible duplicate issues:

https://github.com/actions/checkout/issues/955
https://github.com/actions/runner/issues/1676 (Where I believe this should actually be solved, as this is where the workspace-related code is - see https://github.com/actions/runner/blob/main/src/Runner.Worker/TrackingConfig.cs#L20)
https://github.com/orgs/community/discussions/118587

martiGIT:
any update with that? its still unresolved, what is blocking merging this feature?

Pospelove:
Biba

Pospelove:
Mergeeeeeeeee

matj-sag:
I would also like to see this feature. In my case it's because I'm running the action in a custom container, and I'd rather run it as the user in the container, and not as root. However, this user can only write to a specific directory, and not /, so the default workspace directory of /__w isn't writeable.

enorthcutt (approved):

PR #1503, opened 2023-10-04
Fix cases where there is a "HEAD" file or folder at the root 

SB4597:
of the repo (or, for Windows, Head, head, etc) by making sure HEAD is not treated as a filename


SB4597:
image

monicadiaz68071978 (approved):

PR #1505, opened 2023-10-05
Document that filter doesn't override fetch-depth

jakebailey:
As noted in https://github.com/actions/checkout/pull/1396#issuecomment-1741638179, if you want to make full use of `filter: blob:none`, you still need to set `fetch-depth: 0`, as the default is still `1` which means you still don't get any history.

Happy to change this to whatever makes sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ When Git 2.18 or higher is not in your PATH, falls back to the REST API to downl
     clean: ''
 
     # Partially clone against a given filter. Overrides sparse-checkout if set.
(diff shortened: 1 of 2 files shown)


jenseng:
Or perhaps if a `filter` is set, `fetch-depth` could then implicitly default to 1? I can't think of a scenario where you'd want a simultaneously shallow AND partial clone.

jakebailey:
I assume you mean "default to 0", but yeah, that was something I thought about too, I just wasn't sure how much of a breaking change was palatable.

PR #1509, opened 2023-10-10
Patch 1 using node16

edosaodaro:
Trying to fix "runs" issue from "using: node12" to "using: node16"


TWiStErRob:
* Has this been done in https://github.com/actions/checkout/pull/689 and also https://github.com/actions/checkout/pull/1436 ?

PR #1524, opened 2023-10-26
Update action.yml

goinst-hub:
This partially fixes #940 by updating the description to match the request. However I'm assuming the runner adds in the `Action` part by truncating the description? cc @bryanmacfarlane


joshmgross:
Please file an issue describing the problem you're trying to solve before opening pull requests.

Issue #1475, opened 2023-09-15
Newly added sparse checkout functionality is not cleaning up after itself


markjm:
Thanks @dscho for contributing #1369 . We have started using it and noticed the following issue

`repo-a` / `.github/workflows/workflow-a.yml`

```yaml
jobs:
  job-1:
    runs-on: internal-runner

- uses: actions/checkout@v3
  with:
    sparse-checkout: |
      .github
      src
```

running just the above works as expected. But consider another workflow in the same repo

`repo-a` / `.github/workflows/workflow-b.yml`

```yaml
jobs:
  job-12:
    runs-on: internal-runner

- uses: actions/checkout@v3
- run: ls ./not-src
```

When running `workflow-a`, then workflow `workflow-b`, `workflow-b` will not find any files in `./not-src`.

This is because workflow `workflow-b` runs in the same _work directory as `workflow-a`. On the first run of `workflow-a`, we have polluted the repo in a way this action doesn't know about (and thus can't clean up).

Notice that, even in `workflow-b`:

`git config --list` will show `core.sparseCheckout true`
`git sparse-checkout list` will still include the listed paths
and a simple
`ls` will only show the sparse directories for `workflow-a`

It seems that - similar to the auth cleanup (and other pre-run steps) this action does - we need to disable sparse checkout in the `pre` steps of this action

Note: this is likely only an issue for self-hosted runners which dont get the same cleaning/re-init process github-hosted runners get

Similar-ish issue to recently reported https://github.com/actions/checkout/commit/47fbe2df0ad0e27efb67a70beac3555f192b062f (for submodules, instead of sparse checkout). @megamanics

dscho:
> this is likely only an issue for self-hosted runners

Indeed.

The working directory's path typically contains the workflow name, e.g. `D:\a\_work\repo-a\repo-a`, but not the workflow name.

A solid work-around would be to scorch the worktree (if it exists), right before the `actions/checkout` step.

A more surgical work-around would be to use a shell script like this:

```yaml
- shell: bash
  run: |
    if test true = "$(git config --type bool core.sparsecheckout)"
    then
      rm "$(git rev-parse --git-path info/sparse-checkout)" &&
      git sparse-checkout disable
    fi
```

markjm:
We are looking to use such a workaround in the interim, but it seems this is something that should be fixed here within actions/checkout, no?

jinhyoo-mp:
This caused CI issues for us so we had to remove sparse checkout

[...]

PR #1598, opened 2024-01-31
Explicitly disable sparse checkout unless asked for

dscho:
When a worktree is reused by `actions/checkout` and the first time sparse checkout was enabled, we need to ensure that the second time it is only a sparse checkout if explicitly asked for. Otherwise we need to disable the sparse checkout so that a full checkout is the outcome of this Action.

This fixes https://github.com/actions/checkout/issues/1475


dscho:
It looks as if the `test-proxy` job [fails even without these here changes](https://github.com/actions/checkout/actions/runs/7727700584/job/21066749731), so I'll call this PR good to go.

dist/index.js
```diff
@@ -1282,7 +1287,10 @@ function getSource(settings) {
                 core.endGroup();
             }
             // Sparse checkout
-            if (settings.sparseCheckout) {
+            if (!settings.sparseCheckout) {
+                yield git.disableSparseCheckout();
+            }
+            else {
                 core.startGroup('Setting up sparse checkout');
```
jww3, line 1294: do you want to pop the `.startGroup` / `.endGroup` calls outside of the if?

Consider `core.startGroup('Applying sparse checkout options');`

dscho: No, because we're not actually setting up a sparse checkout. We're _disabling_ it. But saying that would only make sense if there was a worktree already, otherwise there'd be no sparse checkout to disable.

.github/workflows/test.yml
```diff
@@ -175,7 +190,7 @@ jobs:
   test-proxy:
     runs-on: ubuntu-latest
     container:
-      image: alpine/git:latest
+      image: fdev321123/ubuntu-with-git:latest
```
joshmgross, line 193: We should be very careful about adding in external dependencies to our workflows - https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

If this is something we internally built, could we publish it to ghcr.io within this repository or org?

jww3: Agreed.  This was something @fhammerl provided as a proof-of-concept, but I was just reading up about how to publish packages to ghcr.io and like the idea of hosting this container image at ghcr.io/actions (see [neighboring packages here](https://github.com/orgs/actions/packages?ecosystem=container&visibility=public))

.github/workflows/test.yml
```diff
@@ -172,10 +187,20 @@ jobs:
...
   test-proxy:
     runs-on: ubuntu-latest
     container:
-      image: alpine/git:latest
+      image: ubuntu:latest
+      volumes:
+        - /usr/bin/git:/mapped/bin/git
+        - /usr/bin/git-lfs:/mapped/bin/git-lfs
+        - /usr/bin/git-ftp:/mapped/bin/git-ftp
+        - /usr/bin/git-shell:/mapped/bin/git-shell
```
jww3, line 203: After lots of experimentation, it's clear that (as @joshmgross suggested) a custom image hosted on GitHub is going to be the easiest (most maintainable) path forward, here.

```suggestion
      image: ghcr.io/actions/ubuntu-with-git:latest
```

.github/workflows/test.yml
```diff
@@ -185,6 +210,9 @@ jobs:
     env:
       https_proxy: http://squid-proxy:3128
     steps:
+      - name: Update System Path
+        run:  echo "/mapped/bin" >> $GITHUB_PATH
```
jww3, line 214: This can go away.

.github/workflows/test.yml
```diff
@@ -172,10 +187,20 @@ jobs:
       - name: Verify basic
         run: __test__/verify-basic.sh --archive
 
+  inspect:
+    runs-on: ubuntu-latest
+    steps:
+      - run:  ls -all /usr/bin
```
jww3, line 193: This can go away.

jww3 (approved):

dscho:
> LGTM Should we update https://github.com/actions/checkout/blob/main/action.yml#L65 to describe this behaviour?

I don't think that's necessary: This bug fix ensures the described behavior even for non-ephemeral runners.

PR #1599, opened 2024-01-31
ci: fix `test-proxy`

dscho:
According to https://hub.docker.com/r/ubuntu/squid, the proxy won't be available via http://squid.proxy:3128/ but instead via http://localhost:3128/.

I noticed this when the CI failed for https://github.com/actions/checkout/pull/1598, for reasons unrelated to that PR, so I fixed it.


dscho:
Okay, I cannot figure it out, nor justify to continue working on this.

PR #1616, opened 2024-02-20
Create a Docker Container for Proxy Tests, `test-ubuntu-git`

jww3:
The `test-proxy` CI check has been failing since December 19, 2023.

`test-proxy` depends on Docker Hub container image `alpine/git:latest`.  

```
  test-proxy:
    runs-on: ubuntu-latest
    container:
      image: alpine/git:latest
      options: --dns 127.0.0.1
    services:
      squid-proxy:
        image: ubuntu/squid:latest
        ports:
          - 3128:3128
```

It seems a recent update to `alpine/git` caused this test to start failing.

This PR seeks to establish a new git-enabled Linux container image hosted in the GitHub Container Image registry (ghcr.io).  A subsequent PR will update `test-proxy` to utilize this new image.

By establishing a container image that is defined, maintained, published and versioned via the `actions/checkout` repo, we insulate `actions/checkout` CI from external factors that introduce breaking changes.


.github/workflows/update-test-ubuntu-git.yml
```diff
@@ -0,0 +1,55 @@
...
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Use `docker/metadata-action` to preserve tags and labels that exist on the GHCR.io container image. 
+      # - name: Extract metadata (tags, labels) for Docker
+      #   id: meta
+      #   uses: docker/metadata-action@v5.5.1
+      #   with:
+      #     images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
```
jww3, line 44: I'm not convinced this block is necessary at this point.  I'd like to see what happens without it first.

.github/workflows/update-test-ubuntu-git.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # Use `docker/login-action` to log in to GHCR.io. 
+      # Once published, the packages are scoped to the account defined here.
+      - name: Log in to the ghcr.io container registry
+        uses: docker/login-action@v3.0.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
```
fhammerl, line 37: Is `${{ github.actor }}` always the `actions` org?

jww3: I think it's whomever triggers the workflow, (I can devise a quick test to confirm), but I believe it'll still work as desired.  All the [examples](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages) use `${{ github.actor }}`.

The explanation is a bit murky, but some documentation is [here](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility#about-setting-visibility-and-access-permissions-for-packages)

jww3: [`github.repository_owner` is another option](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), but I wanted to try it this way first and see if it works OK.  

I figure it should help make `auth`-related failures in the workflow more clear.  (e.g. joe triggered the workflow, but joe doesn't have the requisite permissions).

It might also be handy to see who published the image in GHCR.io in the event we need to do some forensics.

jww3: I see other workflows in the `actions` org that also use `github.actor` when pushing to `ghcr.io` including actions/runner.

PR #1627, opened 2024-02-27
doc: clarify the behavior of a pull_request event

### LINKED ISSUE #426: What does "Checkout pull request HEAD commit instead of merge commit" mean?
Hi there!

The readme has a section saying how to "[c]heckout pull request HEAD commit instead of merge commit". It isn't really clear what this means, and I tried looking at the source code and it didn't really make sense to me either.

Could you add some more information about this? What is a "merge commit" in this situation?

Some information that might be useful:
* a table with the default behavior and with the behavior with `ref` set to `${{ github.event.pull_request.head.sha }}`
* an example commit DAG

My experience with CI has led me to believe that there's just one way to check out a pull request (the most recent commit on the branch at the time the PR was initiated) so when I came across this option I was really surprised.

cjc7373:
fixes #426

May be the words can be improved. But I think it's necessary to drop a kind notice in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,8 @@ Please refer to the [release page](https://github.com/actions/checkout/releases/
 
 ## Checkout pull request HEAD commit instead of merge commit
 
+The default behavior of a `pull_request` event is to checkout to a ref that already merged your branch into the base branch (So don't be surprised to see the SHA checked out doesn't match your local HEAD SHA). You can change that to checkout your pull request's HEAD by: 
+
 ```yaml
 - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


cory-miller:
I'm going to close this in favor or referring to the documentation page:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

ahans:
I tend to read the docs first before looking through issues to find the answer to some questions. Having this extra sentence in the docs wouldn't hurt at all (and you could in fact even include the link to that GH page there).

PR #1641, opened 2024-03-05
Creating Stale Bot for Checkout repo

ruvceskistefan:


joshmgross:
I would like to get into a better state with the backlog before we enable the stale bot.

ruvceskistefan:
It's just a dry run to see what would stale bot try to stale and close, we can activate it once we're ok with state of the backlog

monicadiaz68071978 (approved):

Issue #1386, opened 2023-06-16
Sparse checkout not working with git version between 2.25 and 2.28


dfdez:
@MasterShihoChief this is an issue that will happen when using git versions between `2.25` and `2.28`being `2.28` the one that should work without any problem, you can see the commit of git [here](https://github.com/git/git/commit/b14ed5adaf87c5943433fd6b1d2cbe8c060f9264)

So basically before `2.28` in order to use `--filter=<filter>` on the `git fetch` the `extensions.partialClone` config had to be set in order to be able to use it...

So here we have 2 solutions:
1. Change the required version to `2.28`
2. Update the current implementation to solve this use case

What do you think about this @dscho?

_Originally posted by @dfdez in https://github.com/actions/checkout/issues/1378#issuecomment-1594880425_

dscho:
My preference would be to bump the minimum Git version.

dfdez:
I think that is going to be the cleaner solution 👍🏽

jkwinder:
This feature isn't working for me today on git 2.41.0 on a windows runner.  This is the error: 
Warning: Unexpected input(s) 'sparse-checkout', valid inputs are ['repository', 'ref', 'token', 'ssh-key', 'ssh-known-hosts', 'ssh-strict', 'persist-credentials', 'path', 'clean', 'fetch-depth', 'lfs', 'submodules', 'set-safe-directory', 'github-server-url']

yml file looks like this:
    - uses: actions/checkout@v3
      with:
        sparse-checkout: |
          .github
          SPA\${{ matrix.angDir }}
          Deployment\SPANexusIQScanAppIdMapping.json

Any ideas what I'm missing?  Or is this still broken on git 2.41?

dscho:
@jkwinder this is a totally different issue than what is described in https://github.com/actions/checkout/issues/1386#issue-1760870334. A new ticket would be more appropriate, and if you choose to open that, do link to a workflow run that demonstrates the issue.

PR #1653, opened 2024-03-12
In consideration of a future v5, update minimum git version.

jww3:


src/git-command-manager.ts
```diff
@@ -11,7 +11,8 @@ import {GitVersion} from './git-version'
 
 // Auth header not supported before 2.9
 // Wire protocol v2 not supported before 2.18
-export const MinimumGitVersion = new GitVersion('2.18')
+// sparse-checkout not supported before 2.25
+export const MinimumGitVersion = new GitVersion('2.25')
```
jww3, line 15: see https://git-scm.com/docs/git-sparse-checkout/2.25.0
and note that https://git-scm.com/docs/git-sparse-checkout/2.24.0 results in a redirect to https://git-scm.com/docs/git-sparse-checkout

src/git-command-manager.ts
```diff
@@ -110,16 +111,7 @@ class GitCommandManager {
...
-
-    // Note, this implementation uses "rev-parse --symbolic-full-name" because the output from
-    // "branch --list" is more difficult when in a detached HEAD state.
-
-    // TODO(https://github.com/actions/checkout/issues/786): this implementation uses
-    // "rev-parse --symbolic-full-name" because there is a bug
-    // in Git 2.18 that causes "rev-parse --symbolic" to output symbolic full names. When
-    // 2.18 is no longer supported, we can switch back to --symbolic.
-
-    const args = ['rev-parse', '--symbolic-full-name']
```
jww3, line 122: TODO:  confirm this `git rev-parse` bug is fixed in git version 2.28+.

src/git-command-manager.ts
jww3: Correction:  minimum git version should be `2.28`  (see #1386)

Vue-Pu (approved):

src/git-command-manager.ts
```diff
@@ -11,7 +11,8 @@ import {GitVersion} from './git-version'
 
 // Auth header not supported before 2.9
 // Wire protocol v2 not supported before 2.18
-export const MinimumGitVersion = new GitVersion('2.18')
+// sparse-checkout not [well-]supported before 2.28 (see https://github.com/actions/checkout/issues/1386)
+export const MinimumGitVersion = new GitVersion('2.28')
```
jww3, line 15: Consider an even higher version in light of https://github.com/actions/checkout/issues/1708.

`2.34.1` exhibits the problem reported in #1708.
`2.43.0` works, but there's probably an earlier candidate that would also work.

(Need to go through the git release notes between version range [2.34 - 2.43] and find where they addressed compatibility issues associated with `repositoryformatversion=1`)

Issue #1651, opened 2024-03-12
Last release failing due to sparse-checkout


joseotoro:
```
/usr/bin/git sparse-checkout disable
git: 'sparse-checkout' is not a git command. See 'git --help'.
Error: The process '/usr/bin/git' failed with exit code 1
```

Job template:
```yaml
jobs:
  FlakyTests:
    runs-on: ubuntu-20.04
    container:
      image: python:2.7.18-buster
    name: Tests
    steps:
    - uses: actions/checkout@v4
      with:
        token: ${{ secrets.BOT_TOKEN }}
        submodules: 'recursive'
```

bewinsnw:
Adding detail from my closed bug (also ubuntu-20.04)
```~$ git --version
git version 2.25.1```

**Comment by jww3** [comment]
We are investigating.  As a workaround, you can pin to v4.1.1

**Comment by clausbuchwald** [comment]
it looks like an outdated git version is being used:

```
Run actions/checkout@v4.1.2
[...]
  /usr/bin/git version
  git version 2.20.1
```

bewinsnw:
Indeed... my bad - the step of my build that fails was the windows build which is older: `git version 2.22.0.windows.1`

But https://github.com/actions/checkout/issues/1386 reports that sparse-checkout wasn't working here until 2.28 and there's a check in the code that the min version for sparse is 2.25 already
https://github.com/actions/checkout/blob/8410ad0602e1e429cee44a835ae9f77f654a6694/src/git-command-manager.ts#L610

bewinsnw:
https://github.com/actions/checkout/blob/main/src/git-source-provider.ts#L212 this is where it's failing in ours - if sparseCheckout isn't enabled it calls `disableSparseCheckout()` which runs `git sparse-checkout disable`, boom

fhammerl:
Rolled back from `v4.1.2` to `v4.1.1`: `actions/checkout@v4` and `actions/checkout@v4.1.1` is ‘latest’ again: https://github.com/actions/checkout/releases

credfeto:
Any reason why 

> Rolled back from `v4.1.2` to `v4.1.1`: `actions/checkout@v4` and `actions/checkout@v4.1.1` is ‘latest’ again: https://github.com/actions/checkout/releases

[...]

PR #1656, opened 2024-03-12
Check git version before attempting to disable `sparse-checkout`

jww3:
This is a fix for a [bug](https://github.com/actions/checkout/issues/1651) discovered when we attempted to release v4.1.2.

Some self-hosted runners have older git versions that pre-date git's `sparse-checkout` command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "checkout",
-  "version": "4.1.2",
+  "version": "4.1.3",
   "description": "checkout action",
   "main": "lib/main.js",
   "scripts": {
--- a/src/git-command-manager.ts
+++ b/src/git-command-manager.ts
@@ -13,7 +13,7 @@ import {GitVersion} from './git-version'
 // Wire protocol v2 not supported before 2.18
 // sparse-checkout not supported before 2.25
 export const MinimumGitVersion = new GitVersion('2.18')
-export const MinimumGitSparseCheckoutVersion = new GitVersion('2.25');
+export const MinimumGitSparseCheckoutVersion = new GitVersion('2.25')
 
 export interface IGitCommandManager {
   branchDelete(remote: boolean, branch: string): Promise<void>
--- a/src/git-command-manager.ts
+++ b/src/git-command-manager.ts
@@ -11,9 +11,9 @@ import {GitVersion} from './git-version'
 
 // Auth header not supported before 2.9
 // Wire protocol v2 not supported before 2.18
-// sparse-checkout not supported before 2.25
+// sparse-checkout not [well-]supported before 2.28 (see https://github.com/actions/checkout/issues/1386)
 export const MinimumGitVersion = new GitVersion('2.18')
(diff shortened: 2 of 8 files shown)


src/git-command-manager.ts
```diff
@@ -11,7 +11,9 @@ import {GitVersion} from './git-version'
 
 // Auth header not supported before 2.9
 // Wire protocol v2 not supported before 2.18
+// sparse-checkout not supported before 2.25
 export const MinimumGitVersion = new GitVersion('2.18')
+export const MinimumGitSparseCheckoutVersion = new GitVersion('2.25');
```
bewinsnw: per https://github.com/actions/checkout/issues/1386 should the minimum version not be 2.28?

jww3: Ah, great catch!  Thanks!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "checkout",
-  "version": "4.1.2",
```
fhammerl, line 3: Should we leave this for the release notes PR?

src/git-command-manager.ts
```diff
@@ -11,7 +11,9 @@ import {GitVersion} from './git-version'
 
 // Auth header not supported before 2.9
 // Wire protocol v2 not supported before 2.18
+// sparse-checkout not [well-]supported before 2.28 (see https://github.com/actions/checkout/issues/1386)
 export const MinimumGitVersion = new GitVersion('2.18')
+export const MinimumGitSparseCheckoutVersion = new GitVersion('2.28')
```
fhammerl, line 16: 👍 I like the new var 👍

src/git-command-manager.ts
```diff
@@ -556,23 +564,23 @@ class GitCommandManager {
 
     // Git version
     core.debug('Getting git version')
-    let gitVersion = new GitVersion()
+    this.gitVersion = new GitVersion()
```
fhammerl, line 567: Why make it a `this.property` ?

src/git-command-manager.ts
```diff
@@ -606,16 +614,14 @@ class GitCommandManager {
 
     this.doSparseCheckout = doSparseCheckout
     if (this.doSparseCheckout) {
-      // The `git sparse-checkout` command was introduced in Git v2.25.0
-      const minimumGitSparseCheckoutVersion = new GitVersion('2.25')
-      if (!gitVersion.checkMinimum(minimumGitSparseCheckoutVersion)) {
+      if (!this.gitVersion.checkMinimum(MinimumGitSparseCheckoutVersion)) {
```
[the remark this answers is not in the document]
jww3: Yes and no.  You'd only see this warning if you were opting to enable `sparse-checkout`.  

The bug in 4.1.2 was that we were disabling `sparse-checkout` without a corresponding version sniff.

fhammerl (approved):
LGTM!

dscho:
Sorry for the breakage!

Issue #1242, opened 2023-03-25
GITHUB_SERVER_URL does not preserve URI prefix


Reverier-Xu:
Hi, I'm trying to deploy Gitea and setup its integrated runners which is compatible with GitHub Actions.

The git server is running under a uri prefix:

```
https://example.com/**service_prefix**/<USER>/<REPO>
```

I notice the code you implemented below:

```typescript
// src/url-helper.ts
...
export function getFetchUrl(settings: IGitSourceSettings): string {
  assert.ok(
    settings.repositoryOwner,
    'settings.repositoryOwner must be defined'
  )
  assert.ok(settings.repositoryName, 'settings.repositoryName must be defined')
  const serviceUrl = getServerUrl(settings.githubServerUrl)
  const encodedOwner = encodeURIComponent(settings.repositoryOwner)
  const encodedName = encodeURIComponent(settings.repositoryName)
  if (settings.sshKey) {
    return `git@${serviceUrl.hostname}:${encodedOwner}/${encodedName}.git`
  }

  // "origin" is SCHEME://HOSTNAME[:PORT]
  return `${serviceUrl.origin}/${encodedOwner}/${encodedName}`
}
...
```

it seems that you just construct repo url with `serviceUrl.origin`, the uri prefix has been ignored.

Is there any particular reason for doing so? If not, I'd like to be able to make a pull request to change its behavior.

frankli0324:
this could be useful not only on a custom deployment of git, consider the scenario when a project requires a repo hosted on another server, and needs to check that repo side by side

apnorton:
Can confirm that PR #1305 worked for me in resolving the issue with a gitea instance and a service prefix; now just for it to be merged...

joshmgross:
We're currently only prioritizing issues for GitHub Actions, which is the primary use case for all of our first-party actions. We don't test compatibility with third-party solutions and thus can't maintain or support any features created exclusively for those third-parties.

This action is open source so you're welcome to fork it and modify it however you wish to support these sort of scenarios.

apnorton:
[...]

PR #1664, opened 2024-03-21
getFetchUrl: compose url with full service url

vicamo:
actions/checkout maybe used by Gitea, which allows installing the server instance in a subdirectory to the service root, therefore a full serviceUrl must be used.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -2402,8 +2402,7 @@ function getFetchUrl(settings) {
     if (settings.sshKey) {
         return `git@${serviceUrl.hostname}:${encodedOwner}/${encodedName}.git`;
     }
-    // "origin" is SCHEME://HOSTNAME[:PORT]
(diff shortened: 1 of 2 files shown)


joshmgross:
See https://github.com/actions/checkout/issues/1242#issuecomment-1773409781 for our current stance on supporting uses of this action outside of GitHub Actions.

vicamo:
> See [#1242 (comment)](https://github.com/actions/checkout/issues/1242#issuecomment-1773409781) for our current stance on supporting uses of this action outside of GitHub Actions.

Yes, but why is there an additional `.origin` access if the `github.server_url` should have already given the right url for further access? It's redundant, isn't it?

Saschaym:
vicamo:for-upstream/use-full-serviceUrl

Saschaym:
#1664

PR #1685, opened 2024-04-18
Add SSH user parameter

cory-miller:
https://github.com/github/actions-launch/issues/427

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -45,6 +45,9 @@ inputs:
       and `CheckHostIP=no` to the SSH command line. Use the input `ssh-known-hosts` to
       configure additional hosts.
     default: true
+  ssh-user:
+    description: >
+      The user to use when connecting to the remote SSH host. By default 'git' is used.
   persist-credentials:
     description: 'Whether to configure the token or SSH key with the local git config'
     default: true
--- a/action.yml
+++ b/action.yml
@@ -48,6 +48,7 @@ inputs:
   ssh-user:
     description: >
       The user to use when connecting to the remote SSH host. By default 'git' is used.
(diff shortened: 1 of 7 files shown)


src/url-helper.ts
```diff
@@ -12,7 +12,8 @@ export function getFetchUrl(settings: IGitSourceSettings): string {
   const encodedOwner = encodeURIComponent(settings.repositoryOwner)
   const encodedName = encodeURIComponent(settings.repositoryName)
   if (settings.sshKey) {
-    return `git@${serviceUrl.hostname}:${encodedOwner}/${encodedName}.git`
+    let user = settings.sshUser.length > 0 ? settings.sshUser : 'git'
```
joshmgross, line 15: Not sure if we care too much about `let` vs `const` in this repo, so feel free to ignore

```suggestion
    const user = settings.sshUser.length > 0 ? settings.sshUser : 'git'
```

action.yml
```diff
@@ -45,6 +45,9 @@ inputs:
       and `CheckHostIP=no` to the SSH command line. Use the input `ssh-known-hosts` to
       configure additional hosts.
     default: true
+  ssh-user:
+    description: >
+      The user to use when connecting to the remote SSH host. By default 'git' is used.
```
joshmgross, line 50:
```suggestion
  ssh-user:
    description: >
      The user to use when connecting to the remote SSH host. By default 'git' is used.
    default: git
```

Should we include the default in the `action.yml`?

joshmgross (approved):

joshmgross (approved):

Issue #485, opened 2021-04-26
Release a new version v3 with `persist-credentials` removed or `false` by default


briansmith:
Currently one has to resort to explicitly specifying `persist-credentials: false` to avoid the credentials being persistent. My understanding is that persisting the credentials gives every step in the job that occurs after `actions/checkout@v2` implicit access to the token. This is not what people expect and this leads people to write jobs that expose their repo to more risk than they otherwise would.

I propose the `persist-credentials` feature be removed completely and then v3 be released. Otherwise, if that's not practical, then at least the default should be changed to `false`.

fmg-dave:
+1

eregon:
Agreed this seems a severe security issue, because it means any workflow using `actions/checkout` basically leaks the token to any process/action in that workflow which can just read it from `.git/config`.

@haampie IIUC it is a problem also with no ssh authentication (the default). The GitHub token is given only to this action and maybe a few other actions/* actions ([`default: ${{ github.token }}`](https://github.com/actions/checkout/blob/2541b1294d2704b0964813337f33b291d3f8596b/action.yml#L24) only work for those AFAIK), but is otherwise given to no other action unless done explicitly (like `with: token: ${{ github.token }}/${{ secrets.GITHUB_TOKEN }}`).
The token is not in the environment.

In other words, `actions/checkout` leaks the token to `.git/config`, making it very easy to read for anything running inside the workflow.
If the token was not written to `.git/config`, then I think stealing the GitHub token would require (one of):
* passing the token explicitly to some action in the workflow and that action gets compromised
* actions/checkout gets compromised
* compromise the workflow definition to e.g. print the token

So, depending on whether the token is explcitly passed to some action:
* If yes, then it seems the only safety net is to [set token permissions](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)
* If no, it would be safe by default with this change or with `persist-credentials: false`, regardless of the token permissions. A workaround is to [set token permissions](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)

I guess GitHub sees setting token permissions as the more general solution.
If so, fine, but then the default should be secure and so the default workflow permissions should be just `contents: read`.

mgoltzsche:
+1

briansmith:
I updated the issue title since v3 and v4 have already shipped, so asking for a new v3 doesn't make sense.

haampie:
[...]

PR #1687, opened 2024-04-20
Change the default value of persist-credentials to false

michi-covalent:
Change the default value of persist-credentials setting from true to false to reduce the risk of unintentionally exposing the GITHUB_TOKEN secret.

Fixes: #485


michi-covalent:
i'd like to get this shipped in the next major release.

monicadiaz68071978 (approved):

Issue #1691, opened 2024-04-23
Version 4.1.3 breaks Sentry auto-detection


PodStuart:
When updating to version 4.1.3 our workflows start to fail with the following error:

`ERROR in Sentry CLI Plugin: Unable to determine version. Make sure to include `release` option or use the environment that supports auto-detection https://docs.sentry.io/cli/releases/#creating-releases.`

Our checkout step looks like this:

```
    - name: Checkout
      uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
      with:
        fetch-depth: 0
        filter: tree:0
        ref: ${{ github.event.pull_request.head.sha }}
```

I see a couple other issues reported against version 4.1.3 and perhaps this has the same root cause.

jww3:
Workaround:  pin workflows to `actions/checkout@v4.1.1`

Proposed fix:  https://github.com/actions/checkout/pull/1692

jww3:
We have created [release `v4.1.4`](https://github.com/actions/checkout/releases/tag/v4.1.4) and marked it as pre-release.  Absent any other issues, we plan to promote `v4.1.4` to `v4` and `latest` by end-of-week.

PodStuart:
Thank you. I've tested 4.1.4 and our workflows all pass.

Issue #1690, opened 2024-04-22
Recent update breaks GitVersion


chadgrant:
Previous config:

 ```
     - name: Checkout
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}

      - name: Install GitVersion
        uses: gittools/actions/gitversion/setup@v0
        with:
          versionSpec: "5.x"

      - name: Determine Version
        id: gitversion
        uses: gittools/actions/gitversion/execute@v0
        with:
          useConfigFile: true
```
Gitversion now errors with "unsupported extension name extensions.worktreeconfig"

I was able to fix it with:

```      - name: Checkout
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}

      - name: fix git
        run: |
          git config --unset-all extensions.worktreeconfig || true
          git config --global --list

      - name: Install GitVersion
        uses: gittools/actions/gitversion/setup@v0
        with:
          versionSpec: "5.x"

      - name: Determine Version
        id: gitversion
        uses: gittools/actions/gitversion/execute@v0
        with:
          useConfigFile: true
```

jww3:
Workaround:  pin workflows to `actions/checkout@v4.1.1`

Proposed fix:  https://github.com/actions/checkout/pull/1692

jww3:
We have created [release `v4.1.4`](https://github.com/actions/checkout/releases/tag/v4.1.4) and marked it as pre-release.  Absent any other issues, we plan to promote `v4.1.4` to `v4` and `latest` by end-of-week.

jww3:
`v4` and `latest` now point to [release `v4.1.4`](https://github.com/actions/checkout/releases/tag/v4.1.4)

Issue #1689, opened 2024-04-22
Recent Update to v4.1.3 Causes LibGit2SharpException with Sparse Checkout


skylayer:
Hello,

After the recent update to `actions/checkout@v4.1.3`, we have encountered an issue when handling repositories after a sparse checkout using LibGit2Sharp. The workflow that was functioning correctly with `v4.1.1` started throwing an exception post-update.

Here's the snippet of the workflow configuration that triggers the issue:

```yaml
- name: Checkout repository
  uses: actions/checkout@v4
  with:
    submodules: true
    fetch-depth: 0
    fetch-tags: true
    filter: blob:limit=100k
```

When our program processes the repository with the above checkout command, it results in the following unhandled exception:

```
Unhandled exception. LibGit2Sharp.LibGit2SharpException: unsupported extension name extensions.worktreeconfig
```

This exception did not occur with version `v4.1.1`. I'm unsure how to proceed and address this issue, and any guidance would be greatly appreciated.

Thank you for looking into this.

jww3:
Hi @skylayer.  We will investigate.  In the meantime, you can pin your workflow to `v4.1.1`

### Example
```
- name: Checkout repository
  uses: actions/checkout@v4.1.1
  with:
    submodules: true
    fetch-depth: 0
    fetch-tags: true
    filter: blob:limit=100k

```

jww3:
Is it possible you're hitting this documented (and recently fixed) lib2git issue?

https://github.com/libgit2/libgit2/issues/6044

TheMoonDawg:
We are also hitting this issue. Primarily when we're utilizing checkout for the `getsentry/action-release@v1` action.

```
 Error: Command failed: /action-release/node_modules/@sentry/cli-linux-x64/bin/sentry-cli releases propose-version
error: Could not automatically determine release name:
	 unsupported extension name extensions.worktreeconfig; class=Repository (6) 

Please ensure your version control system is configured correctly, or provide a release name manually.

Add --log-level=[info|debug] or export SENTRY_LOG_LEVEL=[info|debug] to see more output.
Please attach the full debug log to all bug reports.
```

robdsoule:
We are also seeing the

> We are also hitting this issue. Primarily when we're utilizing checkout for the `getsentry/action-release@v1` action.
> 
> ```
>  Error: Command failed: /action-release/node_modules/@sentry/cli-linux-x64/bin/sentry-cli releases propose-version
> error: Could not automatically determine release name:
[...]

PR #1692, opened 2024-04-23
Disable `extensions.worktreeConfig` when disabling `sparse-checkout`

jww3:
When we released v4.1.3, there were reports that it interfered with some third party tools.

see the following issues:
- https://github.com/actions/checkout/issues/1689
- https://github.com/actions/checkout/issues/1690
- https://github.com/actions/checkout/issues/1691

It turns out that calling `git sparse-checkout disable` has a side effect.  It places the local repo clone in [worktree mode](https://git-scm.com/docs/git-worktree), albeit with only a single "tree".

This fix neutralizes that side effect by explicitly disabling worktree mode.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ on:
       - releases/*
 
 
-# Note that when you see patterns like "ref: test-data/v2/basic" within this workflow, 
+# Note that when you see patterns like "ref: test-data/v2/basic" within this workflow,
 # these refer to "test-data" branches on this actions/checkout repo.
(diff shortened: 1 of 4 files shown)


cory-miller (approved):

huv1k:
When do you plan to release this? I just wonder if we should wait for fix or just pin to old version 🙌

jww3:
@huv1k  We have created [release `v4.1.4`](https://github.com/actions/checkout/releases/tag/v4.1.4) and marked it as pre-release.  Absent any other issues, we plan to promote `v4.1.4` to `v4` and `latest` by end-of-week.

PR #1707, opened 2024-04-25
README: Suggest `user.email` to be `41898282+github-actions[bot]@users.noreply.github.com`

cory-miller:
This is carried over from https://github.com/actions/checkout/pull/1184. Below is the original PR body. This PR updates the base for actions/checkout@v4

---

To get a proper GitHub Actions Bot annotation to a commit one should set the `user.name` and `user.email` to:

```bash
git config user.name "github-actions[bot]"
git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

There are some confusion around this in the `actions/` org itself:

* [**actions/toolkit**/.github/workflows/update-github.yam](https://github.com/actions/toolkit/blob/0db3029fcfcb1393a2b3ccd14caecd7a3460af4e/.github/workflows/update-github.yaml#L24) uses `github-actions[bot]` + `github-actions@github.com` which [does not give a proper avatar on the commit](https://github.com/actions/toolkit/pull/751/commits/d3ad3eeb7f015447f6b9663d1b66995e577776c3)
* [**actions/deploy-pages**/.github/workflows/rebuild-dependabot-prs.yml](https://github.com/actions/deploy-pages/blob/ff669327f73994ba35f45d1cd2fc81d82d1c852c/.github/workflows/rebuild-dependabot-prs.yml#L43-L44) uses the setup suggested in this PR (`github-actions[bot]` + `github-actions[bot]@users.noreply.github.com`) apart from lacking the `41898282+` prefix to the email
* [**actions/runner-images**/.github/workflows/merge_pull_request.yml](https://github.com/actions/runner-images/blob/7fa12b880649ddd0a00fed00b79a5bd4aed462ab/.github/workflows/merge_pull_request.yml#L19-L20) uses `Actions service account` + `no-reply@github.com`
* [**actions/actions-runner-controller**/.github/workflows/update-runners.yaml](https://github.com/actions/actions-runner-controller/blob/9b44f0051ca01556c10e6240f17bb21280be7e39/.github/workflows/update-runners.yaml#L99-L101) uses uses the exact same setup as suggested here, referencing [this discussion](https://github.com/orgs/community/discussions/26560#discussioncomment-3252339) as the source of why that is correct

(Another interesting aspect of the linked to workflows is that some use `--local` when setting the name and email, some use `--global` and some does what is suggested here and use neither of `--local` and `--global`)

Of note is also a third approach from the community, where it strives to achieve custom bot users (+ to get around the limitation of workflow created PR:s not running workflows): https://github.com/wow-actions/use-app-token Such a `env.BOT_NAME` would be quite nice to have built in to use.

An example generated with this setup can be found here: https://github.com/SocketDev/socket-sdk-js/pull/72/commits/24f05748ebcb527cf35dfa5c4cbc2b66d69a45f3

Generated with: https://github.com/SocketDev/workflows/blob/24190c5ca3aa54b086ff1666ed4486f6bc1e9ec9/.github/workflows/reusable-sync.yml#L118-L119

**EDIT:** Swapped the recommendation to include the prefix `41898282+` as I found out that that's the email used when the bot does eg. squash commits and thus its otherwise showing up as being two different users, which looks a bit confusing. See eg: https://github.com/SocketDev/socket-sdk-js/commit/b0948d0da0cfb5b240c69d063cc34a6399abce48


jeffwidman (approved):

README.md
```diff
@@ -279,8 +279,8 @@ jobs:
       - uses: actions/checkout@v4
       - run: |
           date > generated.txt
-          git config user.name github-actions
-          git config user.email github-actions@github.com
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
joshmgross, line 283: We should clarify that this configuration is only valid for github.com - other instances like GHES will have a different ID here.

README.md
```diff
@@ -279,8 +279,8 @@ jobs:
       - uses: actions/checkout@v4
       - run: |
           date > generated.txt
-          git config user.name github-actions
-          git config user.email github-actions@github.com
+          git config user.name "github-actions[bot]"
```
cory-miller, line 282:
```suggestion
          # Note: the following account information will not work on GHES
          git config user.name "github-actions[bot]"
```

joshmgross (approved):

Magdihamed (changes requested):
/

README.md
Magdihamed: r/magdi hamed عمل جيد 👏

szepeviktor:
The user ID can be obtained from GitHub's API.
https://api.github.com/users/github-actions%5Bbot%5D

PR #1736, opened 2024-05-18
Create devcontainer.json

rizwan-r-r:
#656 arm


cory-miller:
This is not related to the issue linked. May look into adding a devcontainer in the future but current PR does not add any useful functionality.

PR #1754, opened 2024-05-27
Use credentials store instead of extraheader

### LINKED ISSUE #162: Auth in extraheader makes it difficult to override
### Description
Since `v2` auth is being persisted in git config, which is great, and makes it much easier to perform git commands. However, persisting the auth in a git `extraheader` config option makes it difficult to override. My particular use case is an [action I maintain that creates pull requests](https://github.com/peter-evans/create-pull-request) for changes made during the workflow. The relevant part of which is that the action accepts a token passed to it and uses it to push to a branch. That token may, or may not, be the same token that was used to checkout the repository. Due to this, I must override whatever auth has been set by `actions/checkout` and push using the token provided.

Initially the action was setting the token in the URL like this:
```
git push https://x-access-token:<token>@github.com/owner/my-repo.git
```
However, the `Authorization` header set by `extraheader` take precedence.

Next, I decided to try overriding the `extraheader` in local git config like this:
```
git -c "http.https://github.com/.extraheader=AUTHORIZATION: basic <basic_credential>" push
```
Initially, this seemed to work, but then I discovered that what was actually happening was instead of overriding the `extraheader` git adds *two* `AUTHORIZATION` headers. This actually makes sense and I was mistaken for thinking it would override the existing `extraheader` config option. What is interesting about this is that the main Git service accepts these requests, but git-lfs does not. You can read the discussion I had about this at the git-lfs repository here: https://github.com/git-lfs/git-lfs/issues/4031

So the only option I was left with to override the auth was to unset the `extraheader` when the action starts, and restore the `extraheader` in local config when the action completes. It feels kind of "hacky" and I don't like that I need to mess with config set by `actions/checkout`.

### Possible solution

During the discussion at the git-lfs repository, @bk2204 [suggested that this could be solved with a credential helper](https://github.com/git-lfs/git-lfs/issues/4031#issuecomment-589254543).

So I tested, assuming that `actions/checkout` set auth in a custom credential helper instead of `extraheader`.
```
git config credential.https://github.com/.helper "! f() { echo username=x-access-token; echo password=$GITHUB_TOKEN; };f"
```
This then allows subsequent git calls to override the auth either by setting the token in the URL, or setting an extraheader with the `git -c` option. Both of these methods take precedence and it only falls back to the credential helper when auth hasn't been explicitly set.

I'm sure changing the auth mechanism is the last thing you want to do, but I wondered what are your thoughts about this and whether you think it's worth considering switching to a credential helper for the next major version bump.

### LINKED ISSUE #415: Git LFS doesn't work with GitHub Enterprise Server on Azure
We have found an issue regarding GitHub checkout action v2, Git LFS and GitHub Enterprise Server (on Azure).

We have a very simply workflow which doesn't work:
```
- uses: actions/checkout@v2
  with:
    lfs: 'true'
```
All git lfs request will be with following error: `HTTP/1.1 400 Bad Request` 
By enabling ` GIT_TRACE=1 and GIT_CURL_VERBOSE=1` we see following:
```
> GET /storage/lfs/3/objects/d5c5871801d62c64f453462558c3a4697ac162730e49d48461ce87bafa83684c HTTP/1.1
> Host: ***.westeurope.cloudapp.azure.com
> Authorization: RemoteAuth AAAAAF72C****
> Authorization: Basic * * * * *
> User-Agent: git-lfs/2.12.1 (GitHub; windows amd64; go 1.14.10; git 85b28e06)
....
> HTTP/1.1 400 Bad Request
> Content-Length: 150
> Content-Type: text/html
> Date: Thu, 17 Dec 2020 10:25:15 GMT
> Server: GitHub.com
```
My current understanding:
The GitHub checkout action is using the extraheader option with basic authorization in the local git config.
For some reason the GitHub Enterprise server is also forcing a RemoteAuth authorization header.
So each request to the git lfs server will contain two headers. 
In this comment https://github.com/git-lfs/git-lfs/issues/4031#issuecomment-589254543 one of the git lfs maintainer mentions that the git lfs server only allows request with **one** authorization header.

At the moment I also don't know exactly why this RemoteAuth authorization header will added.

lehmanju:
Since git's `extraheader` does not support different tokens and has some issue with multiple headers, this PR changes the behavior to use `credential.*.helper`. 
My use case is cloning a repository which has a lot of private submodules from different organizations. Instead of creating a machine user that has access to all organizations I want to use a Github App for that. However, each Github App installation can only see the repositories of its own organization. With multiple organizations, multiple app installations are required, each using a different token.

My solution sets `credential.helper` to `store` and creates a temporary file to which the credentials are written. This option replaces the use of `extraheader` and results in only one authentication header per request.

Additionally, a new option is provided to specify a custom credential helper via `custom-credential-helper`. This setting sets a custom command as git credential helper and can be anything. The setting `useHttpPath` to match against the path of a URL is set to true and the original credential store is used as fallback helper. However, it might be desirable to disable the default helper altogether.

Should 
fix #162
fix #415

I am open for suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ Please refer to the [release page](https://github.com/actions/checkout/releases/
     # Default: true
     persist-credentials: ''
 
+    # Custom git credential helper
+    custom-credential-helper: ''
+
     # Relative path under $GITHUB_WORKSPACE to place the repository
     path: ''
 
--- a/action.yml
+++ b/action.yml
@@ -52,6 +52,8 @@ inputs:
   persist-credentials:
     description: 'Whether to configure the token or SSH key with the local git config'
     default: true
+  custom-credential-helper:
+    description: 'Custom git credential helper'
   path:
     description: 'Relative path under $GITHUB_WORKSPACE to place the repository'
   clean:
--- a/dist/index.js
+++ b/dist/index.js
@@ -166,13 +166,16 @@ class GitAuthHelper {
         this.temporaryHomePath = '';
         this.git = gitCommandManager;
         this.settings = gitSourceSettings || {};
-        // Token auth header
+        this.credentialConfigKey = `credential.helper`;
+        const runnerTemp = process.env['RUNNER_TEMP'] || '';
+        assert.ok(runnerTemp, 'RUNNER_TEMP is not defined');
+        const uniqueId = (0, uuid_1.v4)();
+        this.credentialStorePath = path.join(runnerTemp, `${uniqueId}_credential_store`);
+        this.credentialConfigValue = `store --file ${this.credentialStorePath}`;
         const serverUrl = urlHelper.getServerUrl(this.settings.githubServerUrl);
-        this.tokenConfigKey = `http.${serverUrl.origin}/.extraheader`; // "origin" is SCHEME://HOSTNAME[:PORT]
-        const basicCredential = Buffer.from(`x-access-token:${this.settings.authToken}`, 'utf8').toString('base64');
-        core.setSecret(basicCredential);
-        this.tokenPlaceholderConfigValue = `AUTHORIZATION: basic ***`;
-        this.tokenConfigValue = `AUTHORIZATION: basic ${basicCredential}`;
+        serverUrl.username = `x-access-token`;
+        serverUrl.password = this.settings.authToken;
+        this.tokenCredential = serverUrl.href;
         // Instead of SSH URL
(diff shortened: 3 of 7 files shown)


Mapuppy09 (approved):
Ok , approve  changes

lukins-cz (approved):
Hm 🤔

lehmanju:
This is now ready for a first review. I will cleanup the history afterwards.

lehmanju:
@ericsciple sorry for tagging, but I believe the majority of the code has been written by you. Would you care for a review?

PR #1774, opened 2024-06-12
Check out other refs/* by commit

orhantoy:
The code that chooses what ref to check out assumes that if something is not a branch or a pull request it must be a tag, but that is not necessarily true. When the ref is not a pull request, branch or tag, we should instead check out the commit directly.

### DIFF
--- a/__test__/ref-helper.test.ts
+++ b/__test__/ref-helper.test.ts
@@ -67,6 +67,16 @@ describe('ref-helper tests', () => {
     expect(checkoutInfo.startPoint).toBeFalsy()
   })
 
+  it('getCheckoutInfo refs/', async () => {
+    const checkoutInfo = await refHelper.getCheckoutInfo(
+      git,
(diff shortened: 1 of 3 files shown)


tjgurwara99 (approved):

joshmgross (approved):

cory-miller (approved):

FlorianFranzen:
Please tell me how this example from the tutorial would work in this scenario:
```
on: push
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: |
          date > generated.txt
          # Note: the following account information will not work on GHES
          git config user.name "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
          git add .
          git commit -m "generated"
          git push
```

orhantoy:
@FlorianFranzen That should work but perhaps you're using non-standard ref names as in https://github.com/actions/checkout/issues/1782?

PR #1776, opened 2024-06-12
Pin actions/checkout's own workflows to a known, good, stable version.

jww3:
Dependabot tried to update our internal CI workflows to the bleeding edge.  

It's safer to pin these to a specific version in order to minimize the chances of us having to jump through extra mitigation hoops in the event of a bad release.

In a roll-forward scenario, it's conceivable that a bad release could block the CI of any release candidate intended to replace it.  (It could even potentially block the `update-main-version` workflow -- a *real* chicken-and-egg problem!)  Under those circumstances, we'd have to roll-back to an older version just to get all the necessary workflows passing before we could roll forward.

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v4.1.6
 
       - name: Set Node.js 20.x
         uses: actions/setup-node@v4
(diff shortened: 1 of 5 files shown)


joshmgross (approved):

ratchanonsuttawas (approved):
6ccd57f4c5d15bdc2fef309bd9fb6cc9db2ef1c6

ratchanonsuttawas (approved):

ratchanonsuttawas (changes requested):
+13 -13

ratchanonsuttawas (commented):
>Commit 823fceebd3c828b59f34cc61bc401de251e8897f

PR #1858, opened 2024-08-20
feature: add support for `commit` input

### LINKED ISSUE #1634: Checkout action with branch as ref not consistent through workflow
I recently moved my github actions workflow to trigger on pull_request which has a different checkout behavior than on push. Then to ensure that the git repo was consistent both on push and on pr, I used the checkout action to check out the current branch. 

My main issue with the default checkout on pr was that running git commands doesn't give expected output for instance `git branch` and `git show -s --format=%B`. 

The issue that now occurs is that a workflow starts running on the main branch with one commit, but during the workflow, something else is merged to that branch and the next job then checks out the latest commit instead of the commit it started running with. I tried checkout out the current commit throughout the workflow instead. In python in the workflow a: return Repo(repo_root_path).active_branch.name works on push events, but in prs, results in: ` HEAD is a detached symbolic reference as it points to SHA`. Also with ref: ${{ github.event.pull_request.head.sha }}, but works with ref: current_branch_name

Is there a way to check out a branch at a given commit consistently through a workflow both on push and on pr, for instance:
```
- name: Check out code
              uses: actions/checkout@v4
              with:
                  ref: branch/commit
 ```
 or any other way to check out the same commit throughout a workflow while also having access to the git information?
 uaing fetch-depth:0 makes no difference.

testersen:
Fixes #1634

This pull request adds support for supplying a commit input. This is necessary in order to replicate the standard behavior in pull request synchronization events when not using the default values.

| Commit | Summary |
| :-- | :-- |
| #650ceb0 | Adds a `commit` input in the `action.yaml` file |
| #67b5caa | Modifies `settings.commit` to equal the `commit` input before applying the default value if the input for ref is not found. |
| #8a241b5 | Make `settings.ref` fallback to `settings.commit` if the `commit` input is provided but not `ref` input |
| #267ca9c | Add validation to the `commit` input, to make sure that the input is a string of 40 hexadecimal characters. |
| #3a6c8fb | Added tests to verify the behavior of this pull request. |
| #491fae0 | Result of running `npm run format` |
| #a52fa92 #1be0f94 | Result of running `npm run build` |

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ Please refer to the [release page](https://github.com/actions/checkout/releases/
     # Otherwise, uses the default branch.
     ref: ''
 
+    # The commit SHA to checkout. Used when ref is not specified or is ambiguous. This
+    # can be used as a replacement for ref, or alongside it to checkout a specific
+    # commit of the ref.
+    commit: ''
+
     # Personal access token (PAT) used to fetch the repository. The PAT is configured
     # with the local git config, which enables your scripts to run authenticated git
     # commands. The post-job step removes the PAT.
--- a/__test__/input-helper.test.ts
+++ b/__test__/input-helper.test.ts
@@ -153,7 +153,9 @@ describe('input-helper tests', () => {
     expect(settings.ref).toBeTruthy()
     expect(settings.ref).toStrictEqual('refs/pull/123/merge')
     expect(settings.commit).toBeTruthy()
-    expect(settings.commit).toStrictEqual('0123456789012345678901234567890123456789')
+    expect(settings.commit).toStrictEqual(
+      '0123456789012345678901234567890123456789'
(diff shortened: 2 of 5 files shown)


testersen:
cc @joshmgross - we @TelenorNorgeInternal were hoping to get this merged soon :tm: would you by any chance have time to review sometime before September? Thanks in advance!

monicadiaz68071978 (approved):

monicadiaz68071978 (approved):

mvorisek:
Please let's get this merged.

Our usecase is checkout security hardening. We checkout specific commit SHA, but the SHA can be spoofed from different fork.

So we want to specify branch as `ref` instead and then specific commit using `commit` from this PR.

PR #1869, opened 2024-08-28
Provide explanation for where user email came from

yasonk:
When reading the doc it wasn't clear where the email numerical prefix came from. A short `NOTE` would be helpful.


joshmgross (approved):

joshmgross:
Thanks!

PR #1941, opened 2024-10-11
`url-helper.ts` now leverages well-known environment variables.

jww3:
`url-helper.ts` had unnecessary complexity which could cause it to incorrectly infer common URLs.

This fix leverages [well-known environment variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) for the purposes of determining common URLs.

### DIFF
--- a/__test__/url-helper.test.ts
+++ b/__test__/url-helper.test.ts
@@ -1,0 +1,23 @@
+import * as urlHelper from '../src/url-helper'
+
+describe('isGhes tests', () => {
+  it('basics', async () => {
+    expect(urlHelper.isGhes()).toBeFalsy()
+    expect(urlHelper.isGhes('https://github.com')).toBeFalsy()
+    //expect(urlHelper.isGhes('https://api.github.com')).toBeFalsy()
+    expect(urlHelper.isGhes('https://europe.ghe.com')).toBeFalsy()
+    expect(urlHelper.isGhes('https://test.github.localhost')).toBeFalsy()
+    expect(urlHelper.isGhes('https://src.onpremise.customer.com')).toBeTruthy()
+  })
+})
+
+describe('getServerApiUrl tests', () => {
+  it('basics', async () => {
+    expect(urlHelper.getServerApiUrl()).toBe('https://api.github.com')
+    expect(urlHelper.getServerApiUrl('https://github.com')).toBe('https://api.github.com')
+    expect(urlHelper.getServerApiUrl('https://GitHub.com')).toBe('https://api.github.com')
+    expect(urlHelper.getServerApiUrl('https://europe.ghe.com')).toBe('https://api.europe.ghe.com')
+    expect(urlHelper.getServerApiUrl('https://australia.GHE.COM')).toBe('https://api.australia.ghe.com')
+    expect(urlHelper.getServerApiUrl('https://src.onpremise.customer.com')).toBe('https://src.onpremise.customer.com/api/v3')
+  })
+})
--- a/__test__/url-helper.test.ts
+++ b/__test__/url-helper.test.ts
@@ -1,5 +1,28 @@
 import * as urlHelper from '../src/url-helper'
 
+describe('getServerUrl tests', () => {
+  it('basics', async () => {
+    // Note that URL::toString will append a trailing / when passed just a domain name ...
+    expect(urlHelper.getServerUrl().toString()).toBe('https://github.com/')
+    expect(urlHelper.getServerUrl(' ').toString()).toBe('https://github.com/')
+    expect(urlHelper.getServerUrl('   ').toString()).toBe('https://github.com/')
+    expect(urlHelper.getServerUrl('http://contoso.com').toString()).toBe(
+      'http://contoso.com/'
+    )
+    expect(urlHelper.getServerUrl('https://contoso.com').toString()).toBe(
+      'https://contoso.com/'
+    )
+    expect(urlHelper.getServerUrl('https://contoso.com/').toString()).toBe(
+      'https://contoso.com/'
+    )
+
+    // ... but can't make that same assumption when passed an URL that includes some deeper path.
+    expect(urlHelper.getServerUrl('https://contoso.com/a/b').toString()).toBe(
+      'https://contoso.com/a/b'
+    )
+  })
+})
+
 describe('isGhes tests', () => {
   it('basics', async () => {
     expect(urlHelper.isGhes()).toBeFalsy()
--- a/__test__/url-helper.test.ts
+++ b/__test__/url-helper.test.ts
@@ -4,10 +4,9 @@ describe('isGhes tests', () => {
   it('basics', async () => {
     expect(urlHelper.isGhes()).toBeFalsy()
     expect(urlHelper.isGhes('https://github.com')).toBeFalsy()
-    //expect(urlHelper.isGhes('https://api.github.com')).toBeFalsy()
     expect(urlHelper.isGhes('https://contoso.ghe.com')).toBeFalsy()
     expect(urlHelper.isGhes('https://test.github.localhost')).toBeFalsy()
-    expect(urlHelper.isGhes('https://src.onpremise.customer.com')).toBeTruthy()
+    expect(urlHelper.isGhes('https://src.onpremise.fabrikam.com')).toBeTruthy()
   })
 })
 
@@ -27,7 +26,7 @@ describe('getServerApiUrl tests', () => {
       'https://api.fabrikam.ghe.com'
     )
     expect(
-      urlHelper.getServerApiUrl('https://src.onpremise.customer.com')
-    ).toBe('https://src.onpremise.customer.com/api/v3')
+      urlHelper.getServerApiUrl('https://src.onpremise.fabrikam.com')
+    ).toBe('https://src.onpremise.fabrikam.com/api/v3')
   })
 })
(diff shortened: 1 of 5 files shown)


src/ref-helper.ts
```diff
@@ -249,7 +249,7 @@ export async function checkCommitInfo(
         `Expected head sha ${expectedHeadSha}; actual head sha ${actualHeadSha}`
       )
       const octokit = github.getOctokit(token, {
-        baseUrl: getServerApiUrl(baseUrl),
+        baseUrl: getServerApiUrl(),
```
jww3, line 252: (same as above)  Biggest open question here: is there a **valid** use case where `baseUrl` might represent some other domain?

src/url-helper.ts
```diff
@@ -28,19 +28,19 @@ export function getServerUrl(url?: string): URL {
...
-  }
-
-  return apiUrl
+export function getServerApiUrl(): string {
+  return process.env['GITHUB_API_URL'] || 'https://api.github.com'
 }
 
-export function isGhes(url?: string): boolean {
-  const ghUrl = getServerUrl(url)
+export function isGhes(): boolean {
```
jww3, line 35: This logic now matches what we have in [actions-toolkit](https://github.com/actions/toolkit/blob/ee93b05ee9979c0f7c22e6b648d9fb11b024eeb2/packages/artifact/src/internal/shared/config.ts#L26).

src/github-api-helper.ts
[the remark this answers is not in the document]
TingluoHuang: i think this was introduced to allow checkout to clone repo from a different GHES instance.
ex: workflow runs on GHES-1 and want to clone repo from GHES-2.

There is an input for this use case.

```
    # The base URL for the GitHub instance that you are trying to clone from, will use
    # environment defaults to fetch from the same instance that the workflow is
    # running from unless specified. Example URLs are https://github.com or
    # https://my-ghes-server.example.com
    github-server-url: ''
```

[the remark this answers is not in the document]
easyt: Interesting use case. Yeah, let's just update the `isGhes` function then, and leave the rest alone.

__test__/url-helper.test.ts
```diff
@@ -0,0 +1,23 @@
+import * as urlHelper from '../src/url-helper'
+
+describe('isGhes tests', () => {
+  it('basics', async () => {
+    expect(urlHelper.isGhes()).toBeFalsy()
+    expect(urlHelper.isGhes('https://github.com')).toBeFalsy()
+    //expect(urlHelper.isGhes('https://api.github.com')).toBeFalsy()
```
jww3, line 7: This assertion currently fails when uncommented.

src/url-helper.ts
```diff
@@ -21,26 +21,55 @@ export function getFetchUrl(settings: IGitSourceSettings): string {
...
 }
 
 export function isGhes(url?: string): boolean {
-  const ghUrl = getServerUrl(url)
+  const ghUrl = new URL(
+    url || process.env['GITHUB_SERVER_URL'] || 'https://github.com'
+  )
+
+  const hostname = ghUrl.hostname.trimEnd().toUpperCase()
+  const isGitHubHost = hostname === 'GITHUB.COM'
```
jww3, line 53: should this check also be a `.endsWith` ?  It would allow me to uncomment the disabled unit test above.

__test__/url-helper.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(urlHelper.isGhes('https://europe.ghe.com')).toBeFalsy()
```
easyt, line 8: Swapping with something that more closely resembles a tenant.

```suggestion
    expect(urlHelper.isGhes('https://contoso.ghe.com')).toBeFalsy()
```

src/url-helper.ts
easyt: No. This should only be `github.com`, not a subdomain of it.

__test__/url-helper.test.ts
easyt: You can remove this test case as the URL should always be the base URL and not include a subdomain of `github.com`.

__test__/url-helper.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  })
+})
+
+describe('getServerApiUrl tests', () => {
+  it('basics', async () => {
+    expect(urlHelper.getServerApiUrl()).toBe('https://api.github.com')
+    expect(urlHelper.getServerApiUrl('https://github.com')).toBe('https://api.github.com')
+    expect(urlHelper.getServerApiUrl('https://GitHub.com')).toBe('https://api.github.com')
+    expect(urlHelper.getServerApiUrl('https://europe.ghe.com')).toBe('https://api.europe.ghe.com')
+    expect(urlHelper.getServerApiUrl('https://australia.GHE.COM')).toBe('https://api.australia.ghe.com')
```
easyt:
```suggestion
    expect(urlHelper.getServerApiUrl('https://contoso.ghe.com')).toBe('https://api.contoso.ghe.com')
    expect(urlHelper.getServerApiUrl('https://fabrikam.GHE.COM')).toBe('https://api.fabrikam.ghe.com')
```

src/github-api-helper.ts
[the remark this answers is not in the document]
jww3: Fixed.

easyt (approved):

src/url-helper.ts
```diff
@@ -21,26 +21,56 @@ export function getFetchUrl(settings: IGitSourceSettings): string {
...
+  const isGheHost = hostname.endsWith('.GHE.COM')
```
joshmgross, line 54: Should we use the full public name for this?

https://github.com/newsroom/press-releases/data-residency-in-the-eu

jww3: fixed

src/url-helper.ts
```diff
@@ -21,26 +21,56 @@ export function getFetchUrl(settings: IGitSourceSettings): string {
...
+}
+
+function hasContent(
+  text: string | undefined,
+  allowPureWhitespace: boolean
+): boolean {
+  let refinedText = text ?? ''
+  if (!allowPureWhitespace) {
+    refinedText = refinedText.trim()
+  }
```
joshmgross, line 74: It's not clear to me what this option covers - could you add a test for this scenario?

src/url-helper.ts
```diff
@@ -21,26 +21,56 @@ export function getFetchUrl(settings: IGitSourceSettings): string {
...
+  if (hasContent(url, false)) {
+    resolvedUrl = url!
+  }
+
+  return new URL(resolvedUrl)
 }
 
 export function getServerApiUrl(url?: string): string {
-  let apiUrl = 'https://api.github.com'
+  if (hasContent(url, false)) {
```
joshmgross, line 33: Style nit: `hasContent` here is a bit confusing and I don't think it adds much value.

The checks for `hasContent(url, false)` (where  `allowPureWhitespace` is false) are relatively simple and could be inlined to be a bit more readable.

`if (hasContent(url, false)) {` is equivalent to `if (url) {`, the latter does not require the reader to go figure out what's being set to `false` and knowing how `hasContent` works.

If we removed these cases, we could simplify `hasContent` to always trim and likely inline that usage as well.

```diff
diff --git a/src/url-helper.ts b/src/url-helper.ts
index 5c4fcbd..27969ca 100644
--- a/src/url-helper.ts
+++ b/src/url-helper.ts
@@ -22,15 +22,15 @@ export function getFetchUrl(settings: IGitSourceSettings): string {
 
 export function getServerUrl(url?: string): URL {
   let resolvedUrl = process.env['GITHUB_SERVER_URL'] || 'https://github.com'
-  if (hasContent(url, false)) {
-    resolvedUrl = url!
+  if (url) {
+    resolvedUrl = url
   }
 
   return new URL(resolvedUrl)
 }
 
 export function getServerApiUrl(url?: string): string {
-  if (hasContent(url, false)) {
+  if (url) {
     let serverUrl = getServerUrl(url)
     if (isGhes(url)) {
       serverUrl.pathname = 'api/v3'
@@ -58,19 +58,8 @@ export function isGhes(url?: string): boolean {
 }
 
 function pruneSuffix(text: string, suffix: string) {
-  if (hasContent(suffix, true) && text?.endsWith(suffix)) {
+  if (suffix.trim() && text?.endsWith(suffix)) {
     return text.substring(0, text.length - suffix.length)
   }
   return text
 }
-
-function hasContent(
-  text: string | undefined,
-  allowPureWhitespace: boolean
-): boolean {
-  let refinedText = text ?? ''
-  if (!allowPureWhitespace) {
-    refinedText = refinedText.trim()
-  }
-  return refinedText.length > 0
-}

```

joshmgross (approved):
Comments are not blocking, but please take a look 😄

src/url-helper.ts
jww3: | `str` | `if(str)` | `hasContent(str, true)` | `hasContent(str, true)` |
|-:|-:|-:|-:|
| `null` | `false` | `false` | `false` |
| `undefined` | `false` | `false` | `false` |
| `''` | `false` | `false` | `false` |
| `'abc'` | `true` | `true` | `true` |
| `' '` | `true` | `true` | `false` |

jww3: It's the other way 'round.  A string containing pure whitespace is "truthy," so the predominant pattern would be `if(url?.trim()) { ... }`

I found `hasContent(url, bool)` to be more readable.

src/url-helper.ts
jww3: Unfortunately, there's no easy way to unit test an unexported function.

jww3: I'll give it indirect coverage by writing unit tests around `getServerUrl`.

joshmgross (approved):

src/url-helper.ts
jww3: fixed with an enum

PR #2012, opened 2024-12-19
feat(git config): Set default user.name and user.email in git config

### LINKED ISSUE #13: Set git user and email

chouetz:
Hello :wave:

This PR is a (yet another) tentative to have a solution to issue #13 in the `action/checkout` Github Action.
From the last comments on this issue, it seems everybody agreed on such a setting on this action. I've found a PR #584 on this topic opened 3y ago and which didn't applied the last requested suggestions.
So I decided to do an additional proposal with:
- addition of single boolean input to set a default configuration, as proposed [here](https://github.com/actions/checkout/pull/158#issuecomment-724860815)
- update of the configuration as done on PR #584, using the default values of the [suggestion](https://github.com/actions/checkout/pull/584#discussion_r1637365030), to answer the [question](https://github.com/actions/checkout/issues/13#issuecomment-2207006934) raised on the initial issue.

I'll try to apply any suggestion to this PR so that we can close #13.

Thanks in advance

### DIFF
--- a/src/git-source-provider.ts
+++ b/src/git-source-provider.ts
@@ -275,14 +275,21 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
       settings.githubServerUrl
     )
     if (settings.gitUser) {
-      if (!await git.configExists('user.name', true)) {
+      if (!(await git.configExists('user.name', true))) {
         await git.config('user.name', settings.gitUser, true)
       }
-      if (!await git.configExists('user.email', true)) {
-
-        const userId = await githubApiHelper.getUserId(settings.gitUser, settings.authToken, settings.githubServerUrl);
-        await git.config('user.email', `${userId}+${settings.gitUser}@users.noreply.github.com`, true)
-      } 
+      if (!(await git.configExists('user.email', true))) {
+        const userId = await githubApiHelper.getUserId(
+          settings.gitUser,
+          settings.authToken,
+          settings.githubServerUrl
+        )
+        await git.config(
+          'user.email',
+          `${userId}+${settings.gitUser}@users.noreply.github.com`,
+          true
+        )
+      }
     }
   } finally {
     // Remove auth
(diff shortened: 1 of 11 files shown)


mScott224 (approved):

src/git-source-provider.ts
```diff
@@ -274,6 +274,14 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
       settings.githubServerUrl
     )
+    if (settings.configureUser) {
+      if (!await git.configExists('user.name', true)) {
+        await git.config('user.name', 'github-action[bot]', true)
+      }
+      if (!await git.configExists('user.email', true)) {
+        await git.config('user.email', '41898282+github-actions[bot]@users.noreply.github.com', true)
+      } 
+    }
```
dwg, line 284: This configuration would be inaccurate if using a non github-actions token, right?

For example when using [create-github-app-token](https://github.com/actions/create-github-app-token).

Would it make sense to receive the actor slug via input (defaulting to `"github-action[bot]"`) and lookup the id via `octokit.rest.users.getByUsername`?

chouetz: It makes sense, yes, I'll update this asap, thanks

chouetz: @dwg I've update the PR. I wanted to add tests but it does not seems straightforward to do. If you can have a look, thanks in advance

chouetz:
@joshmgross as it seems you are a maintainer here could you please have a look at this contribution?
Thanks a lot in advance

monicadiaz68071978 (approved):

Issue #862, opened 2022-07-22
Add a 'clean-exclude' option


dapirian:
Currently, the checkout action will delete `node_modules` with `git clean -ffdx`. Since I use (mostly) persistent self-hosted runners, it's extremely inefficient to constantly be deleting and recreating `node_modules`. It takes about a minute just to do the deletion in `git clean` and another 3 minutes to repopulate `node_modules` after that.

It would be much much faster to be able to pass the exclude flag to clean, for example:

```bash
git clean -ffdxe '/node_modules/'

# Or, more explicit...
git clean -ffdx --exclude='/node_modules/'
```

This could just be another argument to the action, so usage might be...
```yaml
- name: Checkout
   uses: actions/checkout@v3
   with:
     clean-exclude: /node_modules/
```

DarrinHidef:
As further clarification, it is handy to have the git clean run, so you know the workspace matches the repo. But for some builds, we don't have control of where the build artifacts get dropped (e.g. with Unity, it produces a Library directory at the top level of the workspace). And those can be expensive to reproduce each time. So we'd like to have their deletion be controlled with a flag, like this. So we can sometimes do clean builds, and sometimes not.

diroussel:
This can be easily done in gitlab by setting GIT_CLEAN_FLAGS to override the args passed to `git clean`.  But in github actions, the work around is very tedious.  I have to set `clean: false` on every checkout of every job, and then add another step to manually do the correct git clean.  But that doesn't work without adding a 'safe.directory'.  It would be so much easier if a custom `clean-exclude` attribute, or env var was supported. 

One way to simplify that I tried was to create a bash script action (we have others in the same repo, in the .github folder) that does the checkout and the git setup, and the clean, but all in one place.  But this doesn't work, as you can't reference other actions in the same repo until you've done the checkout.  So either you need a separate repo, or to do a lot of duplication.

AdarshHatkar:
any update?

MostefaKamalLala:
@dapirian did you come up with a solution?

taco-chen:
Looking forward to have this feature!

bonaert:
@dapirian Created the following PR that adds a `exclude_from_clean` option. See: https://github.com/actions/checkout/pull/2141

PR #2141, opened 2025-04-10
Add `exclude_from_clean` option

bonaert:
See: https://github.com/actions/checkout/issues/862

Adds a `exclude_from_clean` option, which exclude the specified path from the `clean` operation. More concretely, if a value is provided, the action runs `git clean -ffdx -e <exclude_from_clean>`  instead of running `git clean -ffdx`.

This is particularly useful for self-hosted runners, where deleting and re-creating certain directories can sometimes be inneficient and costly (examples: `node_modules`, git LFS files)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 [![Build and Test](https://github.com/actions/checkout/actions/workflows/test.yml/badge.svg)](https://github.com/actions/checkout/actions/workflows/test.yml)
 
+Changes compared to https://github.com/actions/checkout:
+- Can exclude path when cleaning repository (see `exclude_from_clean` argument)
+
 # Checkout V4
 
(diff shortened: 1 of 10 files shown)


BoomchainLabs (approved):

chtaili216:
gh pr checkout 2141

AuntyEms (approved):

monicadiaz68071978 (approved):

kvz:
This is my workaround:

```yaml
    steps:
      - uses: actions/checkout@v4
        with:
          clean: false

      - name: Git Clean
        # While we use actions/checkout@v4 does not have a clean-exclude option,
        # we disable the default clean behavior and do it ourselves.
        # https://github.com/actions/checkout/issues/862
        run: ./_scripts/gitclean.sh
```

On our self-hosted runners where we need some directories to persist. But would love to have this natively supported 👐

PR #2226, opened 2025-07-23
Update actions checkout to use node 24

salmanmkc:
Node 24 support is available on runner version 2.327.1 https://github.com/actions/runner/releases/tag/v2.327.1

Checkout working

Latest main from my branch/fork showing checkout working on node 24
Node 24 action working

PR #2273, opened 2025-09-23
Build  for node 24

diroussel:
Now that node24 is the min runtime, we can target a more recent version of ecmascript, and more recent libs.

Now that https://github.com/actions/checkout/pull/2226 is merged, we can rebuild the dist with less shims.


diroussel:
@salmanmkc I've tested this on my own pipe lines, and it runs just like the v5 tag.

Issue #2275, opened 2025-09-24
Replace uuid with crypto.randomUUID()


trivikr:
Replace uuid with crypto.randomUUID()

Example https://github.com/actions/checkout/blob/ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493/src/github-api-helper.ts#L9

crypto was added globally as follows:
* v23.0.0 | No longer experimental.
* v19.0.0 | No longer behind --experimental-global-webcrypto CLI flag.
* v17.6.0, v16.15.0 |  Added in: v17.6.0, v16.15.0

Docs: https://nodejs.org/api/globals.html#crypto

mohannadmir:
#562 
#2117

PR #2276, opened 2025-09-24
Replace uuid with crypto.randomUUID()

trivikr:
Fixes: https://github.com/actions/checkout/issues/2275


trivikr:
cc @ericsciple @salmanmkc for review

PR #2314, opened 2025-11-24
Update all references from v5 and v4 to v6

ericsciple:
- Updated README.md examples to reference @v6
- Updated all workflow files to use actions/checkout@v6


ng8yb95xhs-byte:
c2d88d3ecc89a9ef08eebf45d9637801dcee7eb5

PR #2336, opened 2025-12-04
Use realpath when including persisted credentials

kwbauson:
This uses the realpath of the git working directory when configuring credentials so that if the checkout is into a symlink, `git` when in the real path still includes them.

I ran into this as an issue on a self-hosted macOS runner whose working directory was under `/var`. On macOS `/var` is a symlink to `/private/var`, so when actions/checkout was configuring credentials it was using the symlink path, which was causing the `includeIf.gitdir` check to fail.


JSTONE1111 (approved):

greg-openai:
Stale pull request message

PR #1180, opened 2023-02-16
Add Ref and Commit outputs

lucacome:
Adds outputs for `ref` and `commit`.

Closes #281

PR #2341, opened 2025-12-08
Add path output

caugner:
Adds an output for `path`.

Similar to https://github.com/actions/checkout/pull/1180, which added outputs for `ref` and `commit`.

Having the checkout path available as an output is especially useful when using the `path` input, and avoids having to duplicate the path manually.

Example:

```diff
 jobs:
   test:
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout content
         id: content
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           repository: mdn/content
           path: mdn/content
           persist-credentials: false

       - name: Run fred
-        working-directory: mdn/content
+        working-directory: ${{ steps.content.outputs.path }}
         env:
-          CONTENT_ROOT: ${{ github.workspace }}/mdn/content/files
+          CONTENT_ROOT: ${{ steps.content.outputs.path }}/files
         run: npm start
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -334,6 +334,7 @@ jobs:
         run: |
           echo "Commit: ${{ steps.checkout.outputs.commit }}"
           echo "Ref: ${{ steps.checkout.outputs.ref }}"
+          echo "Path: ${{ steps.checkout.outputs.path }}"
 
           if [ "${{ steps.checkout.outputs.ref }}" != "test-data/v2/basic" ]; then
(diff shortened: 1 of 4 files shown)


caugner:
@ericsciple Any chance to get this small tested improvement in one of the next releases? We use this action a lot in MDN workflows, and this output would make our life easier. 🙂

caugner:
@ericsciple Just wanted to check if this improvement would be accepted?

Issue #1471, opened 2023-09-13
fetch-tags: true doesn't actually fetch any tags


dataviruset:
After https://github.com/actions/checkout/pull/579 was merged and released with v3.6.0, the `fetch-tags` option was introduced and it's kind of a breaking change for projects that depend on reading tags from the git history. It sets `--no-tags` in the `git fetch` command run by this action. I think this new option should have defaulted to `true` or introduced in v4 instead where it could possibly have had the default `false`.

Nevertheless, using `fetch-tags: true` without overriding the default `fetch-depth: 1` actually doesn't fetch any tags unless the workflow was triggered by a tag push. So I think it's misleading and should be changed to fetch all tags if `fetch-tags` is set to `true`. The change to the `git fetch` command run by this action would be to append `'+refs/tags/*:refs/tags/*'` (`tagsRefSpec`) to the end even though `fetch-depth` is `> 0`.

Related issue: https://github.com/actions/checkout/issues/1467

A possible workaround would be to run another `git fetch` after using this checkout action like in https://github.com/actions/checkout/pull/579#issuecomment-1615799806, but that feels hacky.

Seems the author of the https://github.com/actions/checkout/pull/579 PR also discussed supporting my use-case here: https://github.com/actions/checkout/pull/579#issuecomment-1616149745
I'm sure that this added functionality would be very appreciated. Thanks for your consideration!

tatiesmars:
Up, it would be nice to address this.

ozancaglayan:
This is so annoying and confusing. I set `fetch-depth: 1` and `fetch-tags: true` to express my intention to not checkout all the commits but indeed have all the tags, but no tags are fetched.

5etoolstranslated:
+1

Adding that it should fetch all tags and not only the ones from the `ref` used

zoj613:
Reporting here because I spent a few days trying to figure out why my project's release workflow was failing due to the wheels having the wrong version number. After some digging, it turned out that this change was the cause of this failure since `cibuildwheels` could not use the tag information when buidling wheels. I had to explicitly set `fetch-depth: 0` and `fetch-tags: true` to fix this.

arthurfiorette:
Can confirm its still broken...

`fetch-depth: 0` will be slow for big repositories... I simply added another step to manually fetch tags, which solves this issue perfectly for now.

```yml
- name: Checkout
  uses: actions/checkout@v4

- name: Fetch tags
  run: git fetch --prune --unshallow --tags
```

[...]

Issue #1467, opened 2023-09-06
Can't fetch with `fetch-tags` when triggered by tag


Molkree:
Getting the following error:
`Error: fatal: Cannot fetch both <sha> and refs/tags/<tag> to refs/tags/<tag>`

Example workflow:
```yml
name: Release

on:
  workflow_dispatch:
  # Triggers the workflow on tag events
  push:
    tags: '*'

jobs:
  release:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 250
          fetch-tags: true
```
[Example of a failing run](https://github.com/Molkree/github_actions/actions/runs/6102306304/job/16560501584).

<details>
  <summary>Raw log</summary>
  
```
2023-09-06T20:54:11.5281780Z Requested labels: ubuntu-latest
2023-09-06T20:54:11.5282085Z Job defined at: Molkree/github_actions/.github/workflows/workflow.yml@refs/tags/V1
2023-09-06T20:54:11.5282200Z Waiting for a runner to pick up this job...
2023-09-06T20:54:12.2788574Z Job is waiting for a hosted runner to come online.
2023-09-06T20:54:15.6896729Z Job is about to start running on the hosted runner: GitHub Actions 2 (hosted)
2023-09-06T20:54:17.8444082Z Current runner version: '2.308.0'
2023-09-06T20:54:17.8473502Z ##[group]Operating System
2023-09-06T20:54:17.8474111Z Ubuntu
2023-09-06T20:54:17.8474513Z 22.04.3
2023-09-06T20:54:17.8474765Z LTS
2023-09-06T20:54:17.8475076Z ##[endgroup]
2023-09-06T20:54:17.8475431Z ##[group]Runner Image
2023-09-06T20:54:17.8475752Z Image: ubuntu-22.04
2023-09-06T20:54:17.8476128Z Version: 20230821.1.0
2023-09-06T20:54:17.8476673Z Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20230821.1/images/linux/Ubuntu2204-Readme.md
2023-09-06T20:54:17.8477311Z Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230821.1
2023-09-06T20:54:17.8477802Z ##[endgroup]
2023-09-06T20:54:17.8478168Z ##[group]Runner Image Provisioner
2023-09-06T20:54:17.8478513Z 2.0.266.1
[...]

PR #2356, opened 2026-01-08
Fix tag handling: preserve annotations and explicit fetch-tags

ericsciple:
## Summary

Fixes multiple issues with tag handling in the checkout action.

### 1. `fetch-tags: true` doesn't fetch tags (#1471)
Previously, when using `fetch-tags: true` on a branch checkout, tags were not actually fetched. Now tags are properly included in the fetch refspec.

### 2. Annotated tags converted to lightweight (#290)
When checking out an annotated tag, the tag annotation was lost because the action fetched by commit hash rather than by tag reference. Now tags are fetched via explicit refspec (`+refs/tags/*:refs/tags/*`), which fetches the actual tag objects and preserves annotations.

### 3. Fatal error with tag + `fetch-tags: true` (#1467)
Checking out a tag with `fetch-tags: true` caused a fatal error due to duplicate refspecs. Now only the wildcard refspec is used (the specific tag is redundant when fetching all tags).

### Unit Tests
Added new unit tests.

### E2E Tests

| Scenario | Result |
|----------|--------|
| Tag + fetch-tags: true | ✅ Pass |
| Tag + fetch-tags: false | ✅ Pass |
| Branch + fetch-tags: true | ✅ Pass |
| Branch + fetch-tags: false | ✅ Pass |
| Shallow clone + tags | ✅ Pass |
| Full clone + tags | ✅ Pass |
| Cross-platform (Ubuntu, Windows, macOS) | ✅ Pass |
| Annotated tag moved after trigger | ✅ Fails with clear error |
| Lightweight tag moved after trigger | ✅ Fails with clear error |

### DIFF
--- a/src/git-source-provider.ts
+++ b/src/git-source-provider.ts
@@ -159,7 +159,6 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
     const fetchOptions: {
       filter?: string
       fetchDepth?: number
-      fetchTags?: boolean
       showProgress?: boolean
     } = {}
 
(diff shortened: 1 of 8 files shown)


src/ref-helper.ts
```diff
@@ -170,8 +190,10 @@ export async function testRef(
   // refs/tags/
   else if (upperRef.startsWith('REFS/TAGS/')) {
     const tagName = ref.substring('refs/tags/'.length)
+    // Use ^{commit} to dereference annotated tags to their underlying commit
     return (
-      (await git.tagExists(tagName)) && commit === (await git.revParse(ref))
+      (await git.tagExists(tagName)) &&
+      commit === (await git.revParse(`${ref}^{commit}`))
```
ericsciple: Now that we preserve annotated tags, we need to resolve the commit object not the tag object

src/git-source-provider.ts
```diff
@@ -185,9 +184,22 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
...
-      fetchOptions.fetchTags = settings.fetchTags
-      const refSpec = refHelper.getRefSpec(settings.ref, settings.commit)
+      const refSpec = refHelper.getRefSpec(
+        settings.ref,
+        settings.commit,
+        settings.fetchTags
+      )
       await git.fetch(refSpec, fetchOptions)
+
+      // For tags, verify the ref still points to the expected commit.
```
ericsciple: Since tags are preserved now, we need to validate the tag hasn't moved

TingluoHuang (approved):

ben-manes:
@ericsciple This is fixed in v6.0.2 but that is marked as a pre-release. Was that a mistake and, if not, why not mark it as -beta? This is causing action updates tools to skip/flag it, so I wanted to make sure you were aware of that release marker.

ericsciple:
@ben-manes thanks! I just now removed the pre-release flag (I had originally planned to remove it sooner).

I will watch for issues, then bump the `v6` tag to match next week.

Deejae69:
Cool

Lilili14333:
azure-sdk:increment-package-version-ai-6043413

shanta4100 (commented):
Thank you

PR #2367, opened 2026-01-30
Also rewrite `ssh://` URLs

### LINKED ISSUE #2178: URLs in `.gitmodules` starting with `ssh://` do not work.
You'll need to use an SSH key instead of a PAT if any URL in your `.gitmodules` file starts with `ssh://` instead of `https://` or `git@github.com:` otherwise GitHub will fail to clone the submodule in question with a `Host key verification failed` error.

herrernst:
Like `git@server:` remotes, also rewrite `ssh://git@server/`

Closes: #2178


herrernst:
I've tested it with our many small repos that reference each other by `ssh://` URLs (for a reason), it works now compared to before

greg-openai:
Stale pull request message

PR #2379, opened 2026-02-25
Suppress verbose `git checkout --detached` output with --quiet

maschwenk:
## Summary

When `actions/checkout` runs against a large repository, `git checkout --detach` outputs every file being written to the log. 

image

image

For repos with thousands of files this floods the GitHub Actions log. For our repo this was something like 80k log lines. Why does this matter? Because when Agents read Github Actions logs (which are already chatty) they can easily consume 80k lines of noise for no reason. There are methods for avoiding that, but this output also seems low-value as-is.

In this PR:

I initially was just going to target `checkoutDetach` but then realized there was an existing `progress` config. I figured maybe a more holistic approach would be to reuse this concept. So I have:

- Replaced `--progress` with `--quiet` in `checkout()` to suppress per-file output during checkout
- Add `--quiet` to `checkoutDetach()` for the same reason
- Thread `showProgress` through `checkout()` and `checkoutDetach()` so callers can still opt into `--progress` output; default remains `--quiet`
- Fix `fetchOptions.showProgress` not being set in `git-source-provider.ts` (was always `undefined`, so `--progress` was never applied to fetch either)
- Broaden `show-progress` description from fetch-specific to all git operations

Users can set `show-progress: true` (the default) to keep verbose output, or `show-progress: false` to quiet all git operations.

## **Alternatively, can keep this very directly constrained to the detached checkout**

(note to self, do not rebase this branch as we are referencing it a specific commit in it already)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -78,7 +78,7 @@ inputs:
     description: 'Whether to fetch tags, even if fetch-depth > 0.'
     default: false
   show-progress:
-    description: 'Whether to show progress status output when fetching.'
+    description: 'Whether to show progress status output for git operations.'
     default: true
   lfs:
     description: 'Whether to download Git-LFS files'
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ Please refer to the [release page](https://github.com/actions/checkout/releases/
     # Default: false
     fetch-tags: ''
 
-    # Whether to show progress status output when fetching.
+    # Whether to show progress status output for git operations.
     # Default: true
     show-progress: ''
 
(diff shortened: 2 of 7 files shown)


nagypeterjob:
good stuff, +1 for merging this 👍🏻

action.yml
```diff
@@ -78,7 +78,7 @@ inputs:
     description: 'Whether to fetch tags, even if fetch-depth > 0.'
     default: false
   show-progress:
-    description: 'Whether to show progress status output when fetching.'
+    description: 'Whether to show progress status output for git operations.'
```
maschwenk, line 81: `show-progress` does not really imply this is exclusively related to `fetch`. Only this little helptext.

PR #2428, opened 2026-05-11
Handle checkout 500 retries with backoff

pillarsdotnet:
## Summary

I've been getting intermittent build failures in my GitHub Actions pipeline due to ephemeral GitHub 500-series server failures. I'd like to harden my workflows to retry these errors instead of failing immediately. Most such failures happen when calling actions/checkout. Therefore, this PR makes the following changes:

- switch the shared retry helper to exponential backoff for retry attempts
- cover the retry timing in unit tests, including transient 500-style failures
- regenerate the bundled dist output


YunWu955 (approved):

satmandu:
OMFG. Thank you for this.

PR #2454, opened 2026-06-12
block checking out fork pr for pull_request_target and workflow_run

aiqiaoy:
Implements https://github.com/github/c2c-actions/pull/10159

TL;DR; This PR adds a check that refuses to check out fork pull request code when the workflow trigger is either `pull_request_target` or `workflow_run`, unless the workflow author explicitly opts in via a new input `allow-unsafe-pr-checkout: true`.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -2834,7 +2834,7 @@ function assertSafePrCheckout(input) {
         `This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch ` +
         `cache scope, and runner access. Fetching and executing a fork's code in that trusted ` +
         `context commonly leads to "pwn request" vulnerabilities. To opt in after reviewing ` +
-        `the risks at https://gh.io/securely-using-pull-request-checkout, set ` +
+        `the risks at https://gh.io/securely-using-pull_request_target, set ` +
         `'allow-unsafe-pr-checkout: true' on the actions/checkout step.`);
 }
 function pushIfSha(target, value) {
--- a/dist/index.js
+++ b/dist/index.js
@@ -2793,9 +2793,11 @@ function assertSafePrCheckout(input) {
         return;
     }
     let prHeadRepoId;
+    let prHeadRepoFullName;
     const prShas = [];
     if (eventName === 'pull_request_target') {
         prHeadRepoId = (0, ref_helper_1.fromPayload)('pull_request.head.repo.id');
+        prHeadRepoFullName = (0, ref_helper_1.fromPayload)('pull_request.head.repo.full_name');
         pushIfSha(prShas, (0, ref_helper_1.fromPayload)('pull_request.head.sha'));
         pushIfSha(prShas, (0, ref_helper_1.fromPayload)('pull_request.merge_commit_sha'));
     }
@@ -2805,7 +2807,13 @@ function assertSafePrCheckout(input) {
             return;
         }
         prHeadRepoId = (0, ref_helper_1.fromPayload)('workflow_run.head_repository.id');
+        prHeadRepoFullName = (0, ref_helper_1.fromPayload)('workflow_run.head_repository.full_name');
         pushIfSha(prShas, (0, ref_helper_1.fromPayload)('workflow_run.head_commit.id'));
+        // For `pull_request_target`-triggered workflow_run, `head_sha` is the base
+        // default branch SHA (not the PR head)
+        if (wrEvent !== 'pull_request_target') {
+            pushIfSha(prShas, (0, ref_helper_1.fromPayload)('workflow_run.head_sha'));
+        }
     }
     // (A) Fork PR?
     if (typeof prHeadRepoId !== 'number' || prHeadRepoId === baseRepoId) {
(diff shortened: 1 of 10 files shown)


dist/index.js
```diff
@@ -2144,6 +2145,17 @@ function getInputs() {
         // Determine the GitHub URL that the repository is being hosted from
         result.githubServerUrl = core.getInput('github-server-url');
         core.debug(`GitHub Host URL = ${result.githubServerUrl}`);
+        // Allow unsafe PR checkout (opt-in for pull_request_target / workflow_run fork PRs)
```
AdnaneKhan, line 2148: Can this apply to merge/head ref + issue_comment constructions by default as well?

aiqiaoy: Thanks for the review, The risk of `issue_comment` is on our radar, however it will not be included in this change. This is because `issue_comment` event payload only has the PR number (no ref or sha), so we can't detect fork PR without an extra API call and `issue_comment` has a slightly different threat model (gated by who can comment rather than who can open a PR) that deserves its own design.

src/unsafe-pr-checkout-helper.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (typeof prHeadRepoId !== 'number' || prHeadRepoId === baseRepoId) {
+    return
+  }
+
+  // (B) We cannot check for all fork PR refs so check to see
+  // if the resolved input points to the fork PR sha we have in the payload
+  const baseQualifiedRepository = `${github.context.repo.owner}/${github.context.repo.repo}`
+  const repositoryDiffersFromBase =
+    input.qualifiedRepository.toLowerCase() !==
+    baseQualifiedRepository.toLowerCase()
```
yaananth: should we compare against head repo as well not just base repo to support checking out unrelated repos in fork?

src/unsafe-pr-checkout-helper.ts
```diff
@@ -0,0 +1,80 @@
...
+    prHeadRepoId = fromPayload('pull_request.head.repo.id')
+    pushIfSha(prShas, fromPayload('pull_request.head.sha'))
+    pushIfSha(prShas, fromPayload('pull_request.merge_commit_sha'))
+  } else {
+    const wrEvent = fromPayload('workflow_run.event')
+    if (typeof wrEvent !== 'string' || !wrEvent.startsWith('pull_request')) {
+      return
+    }
+    prHeadRepoId = fromPayload('workflow_run.head_repository.id')
+    pushIfSha(prShas, fromPayload('workflow_run.head_commit.id'))
```
yaananth, line 44: can head commit. be ever null? do we also need to add  head_sha in here?

src/unsafe-pr-checkout-helper.ts
aiqiaoy: good catch, updated!

src/unsafe-pr-checkout-helper.ts
aiqiaoy: Not really sure if `head.commit` can be null (never null in my testing). Still added `head_sha` when the original event that triggered `workflow_run` is not `pull_request_target`. This is because when `workflow_run` is triggered by `pull_request_target`, the `head_sha` points to the base branch, not the PR head.

src/unsafe-pr-checkout-helper.ts
```diff
@@ -0,0 +1,80 @@
...
+import {fromPayload} from './ref-helper'
+
+const PR_REF_PATTERN = /^refs\/pull\/[0-9]+\/(?:head|merge)$/
+
+export interface IUnsafePrCheckoutInput {
+  qualifiedRepository: string
+  ref: string
+  commit: string
+  allowUnsafePrCheckout: boolean
+}
```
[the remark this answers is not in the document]
aiqiaoy: Added

src/unsafe-pr-checkout-helper.ts
```diff
@@ -0,0 +1,80 @@
...
+    pushIfSha(prShas, fromPayload('pull_request.head.sha'))
+    pushIfSha(prShas, fromPayload('pull_request.merge_commit_sha'))
+  } else {
+    const wrEvent = fromPayload('workflow_run.event')
+    if (typeof wrEvent !== 'string' || !wrEvent.startsWith('pull_request')) {
+      return
+    }
+    prHeadRepoId = fromPayload('workflow_run.head_repository.id')
+    pushIfSha(prShas, fromPayload('workflow_run.head_commit.id'))
+  }
```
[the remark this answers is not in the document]
aiqiaoy: Added head_sha when the original event that triggered workflow_run is not pull_request_target. This is because when workflow_run is triggered by pull_request_target, the head_sha points to the base branch, not the PR head.

yaananth (dismissed):

yaananth (approved):

fproulx-boostsecurity:
This is awesome @aiqiaoy !

gilescope:
Gold star! That's a fantastic way to help secure the world.

gm122921980-create (commented):
👍

PR #2463, opened 2026-06-16
upgrade module to esm and update dependencies

aiqiaoy:
The latest version of `@actions/http-client`, `@actions/core` and `@actions/github` packages are ESM only. We need to migrate checkout to ESM so we can update our dependencies to get the security fixes. 

Mostly followed https://github.com/actions/upload-artifact/pull/762


src/main.ts
```diff
@@ -1,28 +1,28 @@
...
   try {
     const sourceSettings = await inputHelper.getInputs()
 
     try {
       // Register problem matcher
-      coreCommand.issueCommand(
-        'add-matcher',
-        {},
-        path.join(__dirname, 'problem-matcher.json')
+      core.info(
```
aiqiaoy, line 16: `issuseCommand` is not exported, reaching past export into submodules is supported under CommonJS but not ESM. 

This "core.info(`::add-matcher::${path.join(__dirname, 'problem-matcher.json')}" is functionally equivalent

yaananth (approved):

PR #2508, opened 2026-07-08
Apply filter options to submodule update

### LINKED ISSUE #1714: `filter` options not applied to submodules?
Checking the log of an action which is using the following
```yml
- uses: actions/checkout@v4
  with:
    submodules: true
    fetch-depth: 0
    filter: blob:none
```
The submodule checkout step indicates the filter is not applied
```
# Fetching the primary repo
git -c protocol.version=2 fetch --prune --no-recurse-submodules --filter=blob:none ...
# fetching / checking out submodules
git -c protocol.version=2 submodule update --init --force
```
I was expecting to see the `--filter=...` also present in the submodule. Is there a reason this is not the case?

swt2c:
Fixes: https://github.com/actions/checkout/issues/1714

### DIFF
--- a/__test__/git-command-manager.test.ts
+++ b/__test__/git-command-manager.test.ts
@@ -403,6 +403,72 @@ describe('Test fetchDepth and fetchTags options', () => {
   })
 })
 
+describe('Test submoduleUpdate filter option', () => {
+  beforeEach(async () => {
+    mockFileExistsSync.mockReset()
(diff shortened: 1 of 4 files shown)


hilljessica8787-dot (approved):

swt2c:
@aiqiaoy @yaananth any thoughts here?  This would be a great performance improvement for one of my workflows.

PR #2552, opened 2026-08-06
Fix unmaterialized LFS files on persistent/self-hosted runners

KlotzAndrew:
Problem

On persistent/self-hosted runners that reuse a working directory across jobs, LFS-tracked files can be left as pointer files instead of real content when a lfs: false checkout precedes a lfs: true checkout of the same repo.

Sequence that triggers it:
 1. A lfs: false run sets GIT_LFS_SKIP_SMUDGE=1, so git checkout writes LFS files to the working tree as pointers.
 2. A later lfs: true run on the same worktree fetches the objects into .git/lfs, but git checkout sees the tree already at the target commit and does not re-smudge the already-present pointer files.

Result: the objects are in the local cache, but the working tree still contains pointers. The action relies on the smudge filter firing during git checkout to materialize content - which holds on a clean tree (GitHub-hosted runners) but not on a reused one.

Related: actions/checkout#270


dist/index.js
```diff
@@ -35816,6 +35816,9 @@ class GitCommandManager {
         const output = await this.execGit(['rev-parse', '--symbolic-full-name', '--verify', '--quiet', 'HEAD'], true);
         return !output.stdout.trim().startsWith('refs/heads/');
     }
+    async lfsCheckout() {
+        await this.execGit(['lfs', 'checkout']);
```
thschalch, line 35820: In case the input parameter `submodules` is `true` or `recursive`, I would expect that also the LFS files of the submodules are checked out. Maybe we should also add a command like:

```
git submodule foreach [--recursive] git lfs checkout
```

thschalch:
This PR is related (or almost a duplicate) of #1392.

dist/index.js
KlotzAndrew: makes sense, added a `git.submoduleForeach`

KlotzAndrew:
> This PR is related (or almost a duplicate) of #1392.

Oh you're right, happy to get either in as long as this LFS fix gets merged

@sxlijin any strong feelings? looks like you've been sitting on an unmerged fix for a hot minute

sxlijin:
No strong feelings, happy for either of them to get merged! Step 1 will be getting attention from someone with maintainer permissions though... 😅

PR #2560, opened 2026-08-25
Log Git initialization failures before REST fallback

### LINKED ISSUE #2435: Errors from initializing the gitCommandManager are ignored and never displayed
We were in a situation where one of our runners would fallback to the GitHub REST API. Even though `actions/checkout` was able to find our locally installed git.

This made me rummage through the source code of the action and I noticed that all errors from creating/initializing the CommandManager will be ignored. Even though the errors themselves would have been very useful. For example, we were using sparse checkouts, and apparently the sparse checkout version requires git 2.28. If the errors would have been printen then this wouldn't have taken us a few hours to find out that the locally installed git version was outdated (2.23).

I would like to propose to add a simple print//info on the contents of the error message in the [`getGitCommandManager` function's catch block.](https://github.com/actions/checkout/blob/900f2210b1d28bbbd0bd22d17926b9e224e8f231/src/git-source-provider.ts#L366-L371)

```typescript
    } catch (err) {
    // add this?
    core.info(`Failed to initialize CommandManager: ${err.message}`)
    
    // Git is required for LFS
    if (settings.lfs) {
      throw err
    }
```

This would result in an info print like: `Failed to initialize CommandManager: Minimum Git version required for sparse checkout is 2.28. Your git ('/path/to/git') is 2.23`

Another option would be to also check `settings.sparseCheckout`. And if that setting is set, just like `settings.lfs` to rethrow err. As a valid git version is required for sparseCheckouts and the action is unable to do a sparse checkout when doing a download over the REST API.

mameikagou:
## Summary

- report the Git command manager initialization error before falling back to the REST API
- add a regression test that verifies the diagnostic and preserves the existing fallback behavior
- rebuild `dist/index.js`

This makes version and sparse-checkout failures visible without changing the current LFS or REST fallback semantics.

Closes #2435

## Testing

- `npm test -- --runInBand`
- `npm run format-check`
- `npm run lint`
- `npm run build`


De-ASI-INTERFACE (approved):

src/git-source-provider.ts
```diff
@@ -381,6 +381,9 @@ async function getGitCommandManager(
       settings.sparseCheckout != null
     )
   } catch (err) {
+    const errorMessage = err instanceof Error ? err.message : String(err)
+    core.info(`Failed to initialize CommandManager: ${errorMessage}`)
+
     // Git is required for LFS
     if (settings.lfs) {
```
r2tbcb9n2c-hash: dist/index.js

PR #2568, opened 2026-09-05
Add a quiet input to suppress git output

### LINKED ISSUE #2409: support --quiet
https://github.com/actions/checkout/issues/894 is closed but not fixed
https://github.com/actions/checkout/issues/18 is closed but not fixed

even if [progress is off](https://github.com/actions/checkout/pull/1067) quiet is useful,
does not need to be the default, just give us the option to set it, it's not that hard 😢 

especially with `fetch: 0` there is so much noise

Image

cc @simonbaird @z2oh

mostafazh:
Adds a `quiet` input (default `false`) that passes `--quiet` to the git commands checkout runs: `fetch`, `checkout`, `checkout --detach`, and `submodule update`.

Fixes #2409.

## Why `show-progress: false` is not enough

#1067 made `--progress` conditional, which handles the progress sideband. The noise reported in #2409 is a different thing: the **per-ref fetch summary**. At `fetch-depth: 0` the refspec is `+refs/heads/*:refs/remotes/origin/*` plus `+refs/tags/*:refs/tags/*`, so git prints a `From <url>` line and then one `* [new branch]` / `* [new tag]` line for every ref in the repository. That output is not progress, so no `--progress` toggle removes it. `--quiet` does.

## Evidence

Three jobs on the same runner, all fetching the same repository (`actions/checkout`, 61 branches + 68 tags) at `fetch-depth: 0`. The middle arm is a control — this branch with `quiet: false` — to show the change is inert unless the input is set.

| Arm | `Fetching the repository` group | `* [new ...]` lines | Whole checkout step |
|---|---|---|---|
| stock `actions/checkout@v7` | 135 lines | 131 | 216 lines |
| this branch, `quiet: false` | 135 lines | 131 | 217 lines |
| this branch, `quiet: true` | **3 lines** | **0** | **83 lines** |

The control arm's fetch group is content-identical to the baseline (0 differing lines once timestamps are stripped), so `quiet` is the only variable.

Baseline:

```
##[group]Fetching the repository
[command]/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
From https://github.com/actions/checkout
 * [new branch]      Update-description      -> origin/Update-description
 * [new branch]      copilot/backport-2518-releases-v2 -> origin/copilot/backport-2518-releases-v2
 ... 129 more [new branch] / [new tag] lines ...
##[endgroup]
```

With `quiet: true`:

```
##[group]Fetching the repository
[command]/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --quiet origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
##[endgroup]
```

Full run: https://github.com/mostafazh/checkout/actions/runs/33961567354 — workflow: https://github.com/mostafazh/checkout/blob/quiet-ab-test/.github/workflows/quiet-ab-test.yml

## `--progress` vs `--quiet`

These two drive different output channels and **compose** rather than conflict, so this PR passes both where both apply:

- `--progress` / `--no-progress` controls only the progress meter (`Receiving objects`, `Updating files: N%`).
- `-q` / `--quiet` controls verbosity: the `From <url>` + `* [new ...]` ref summary for `fetch`, `Switched to branch ...` for `checkout`. It also turns the meter off as a side effect.

When both are given, the meter follows `--progress` (git evaluates forced progress before verbosity, and `git-checkout(1)` says `--progress` applies "regardless of `--quiet`"), while everything else stays suppressed by `--quiet`. Measured on git 2.50.1 over `file://` with stderr not a terminal:

| `git fetch` flags | progress lines | ref-summary lines |
|---|---|---|
| *(none)* | 0 | 11 |
| `--progress` | 3 | 11 |
| `--quiet` | 0 | 0 |
| `--quiet --progress` | 3 | 0 |
| `--progress --quiet` | 3 | 0 |

Order is irrelevant; each flag acts on its own channel. So `checkout` keeps its unconditional `--progress` and merely gains `--quiet`, which means a large checkout still reports progress while the chatter goes away.

## Separate observation, not fixed here

While tracing this I noticed `show-progress` currently has no effect. `getSource` declares `showProgress?: boolean` on the `fetchOptions` object in `src/git-source-provider.ts` but never assigns `settings.showProgress` to it, so `options.showProgress` is always `undefined` and `--progress` is never passed to `fetch`. This is visible in the baseline log above: the step's resolved inputs report `show-progress: true`, yet the fetch command carries no `--progress`. It appears to date back to #1067, which added the field to the type but not the assignment.

I have deliberately left it alone to keep this PR focused, and because wiring it up would make the default output noisier for everyone. Happy to open a separate issue or PR if you would like it addressed.

## Notes

- `git lfs fetch` has no `--quiet` flag, so LFS output is unchanged.
- The default is `false`, so existing workflows are unaffected.

## Per CONTRIBUTING

- **Tests added.** `__test__/git-command-manager.test.ts` gains a `Test quiet option` block (11 cases) asserting the exact argv for `fetch`, `checkout` with and without a start point, `checkoutDetach` and `submoduleUpdate` in both the quiet and non-quiet states, plus `--quiet` taking precedence over `--progress` and the omitted-argument default staying non-quiet. `__test__/input-helper.test.ts` covers the input default and parsing.
- `npm run test` — 140 passing
- `npm run format` — clean
- `npm run build` — `dist/index.js` rebuilt and committed


mostafazh:
Before and After Logs: https://gist.github.com/mostafazh/7badb07f545eb6d1913c382962b346cc (48% less in Number of Lines)

mostafazh:
@aiqiaoy @yaananth can you please review? 🙏

